Issue #20, opened 2020-01-31
Extend result factories with field to value mapping methods


johanneshiry:
Currently, the result factories are only able to build a ResultEntity based on a fieldName -> valueMapping which allows reading and creating of POJOs and can be used to read data into the data model from the outside. The other way around e.g. for writing operations is not supported yet. Hence to support writing, the factories need to be extended to allow for a creating a mapping based on a POJO.

PR #21, opened 2020-01-31
Extend result factories with field to value mapping methods 

johanneshiry:
addresses #20 

Instead of extending result factories, a model processor is introduced that allows for model processing to prepare the models to be written.


johanneshiry:
!test

johanneshiry:
!test

johanneshiry:
!test

johanneshiry:
!test

johanneshiry:
!test

src/main/java/edu/ie3/io/factory/EntityFactory.java
```diff
@@ -24,8 +23,9 @@
 
   protected final List<Class<? extends T>> classes;
 
-  public EntityFactory(Class<? extends T>... classes) {
-    this.classes = Arrays.asList(classes);
+  public EntityFactory(Class<? extends T>... allowedClasses) {
+    this.classes = Arrays.asList(allowedClasses);
+    TimeTools.initialize(ZoneId.of("UTC"), Locale.GERMANY, "yyyy-MM-dd HH:mm:ss");
```
sebastian-peter: Do we need this here or does it make more sense in a more universal place?

src/main/java/edu/ie3/io/processor/EntityProcessor.java
```diff
@@ -0,0 +1,214 @@
...
+
+  /**
+   * Determines if the provided class is a subclass of ResultEntity
+   *
+   * @param cls the provided class
+   * @return true if it is a subclass of ResultEntity, false otherwise
+   */
+  private boolean isResultModel(Class<?> cls) {
+    return ResultEntity.class.isAssignableFrom(cls);
+  }
```
sebastian-peter: Shouldn't this be handled in ResultEntityProcessor, i.e. by inheritance?

src/main/java/edu/ie3/io/processor/EntityProcessor.java
```diff
@@ -0,0 +1,214 @@
...
+   * Standard method to process a Quantity to a String based on a method return object
+   *
+   * @param quantity the quantity that should be processed
+   * @param fieldName the field name the quantity is set to
+   * @param resultModel true if the processed model is of type ResultEntity, false otherwise
+   * @return an optional string with the normalized to {@link StandardUnits} value of the quantity
+   *     or empty if an error occurred during processing
+   */
+  protected Optional<String> handleQuantity(
+      Quantity<?> quantity, String fieldName, boolean resultModel) {
```
sebastian-peter: I feel like handleQuantity should just be specialized in ResultEntityProcessor? Then resultModel boolean could be dropped.

src/main/java/edu/ie3/io/processor/EntityProcessor.java
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Register the class provided in the constructor
+   *
+   * @param cls class to be registered
+   * @return an array of strings of all field values of the registered class
+   */
+  private String[] registerClass(Class<?> cls) {
+
+    try {
+      Arrays.stream(Introspector.getBeanInfo(cls, Object.class).getPropertyDescriptors())
```
sebastian-peter: Isn't this the devil's method?

src/main/java/edu/ie3/io/factory/EntityFactory.java
sebastian-peter: (I realize that might be a lot of work)

src/main/java/edu/ie3/io/processor/EntityProcessor.java
johanneshiry: Yep, but we need it here to determine the fields we want to write. In contrast to the application where it causes performance issues, here it is only used once to register the class that should be written out. Hence, it is acceptable as it is a one shot call and not (like in the other application) called for each element to be processed.

src/main/java/edu/ie3/io/processor/EntityProcessor.java
johanneshiry: I agree on this. Restructured the code accordingly and moved the special handling into `ResultEntityProcessor` while keeping the generalized method and avoid code duplicates. Please let me know if you think this is a valid way to go.

src/main/java/edu/ie3/io/processor/EntityProcessor.java
johanneshiry: normally yes, but with the comment below I totally removed it as it is unnecessary now.

src/main/java/edu/ie3/io/factory/EntityFactory.java
johanneshiry: I _assume_ this is for new (sadly) the only way to go with, as this is the convention introduced for the time formatting for i/o operations. In the long term it might make sense to either a) set this pattern to the standard in utils b) add a second TimeTools pattern that can easily be called c) introduce sth. like "PowerSystemDatamodelContansts" class that hols the pattern as the default one.

johanneshiry:
!test

johanneshiry:
!test

PR #35, opened 2020-02-11
Added ResultEntityFactory + set java version to 1.8

johanneshiry:
This PR solves two (discussed but not written down issues):

- set java version back to 1.8 as we primarily use features that are available in java 8 as well (except a few ones) and hence to improve backward compatibility java 8 seems more suitable
- added a ResultEntityFactory as a new abstract class for all ResultEntityFactories implementation so allow the creation of generic ResultEntity readers as well as removing some code duplications

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: trusty
 language: java
 jdk:
-  - oraclejdk11
+  - oraclejdk8
 script:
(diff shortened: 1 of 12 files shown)


johanneshiry:
!test

johanneshiry:
!test

Issue #40, opened 2020-02-13
Extend Operator with static field NO_OPERATOR


johanneshiry:
Currently an `OperatorInput` is required by all input entities, but it might be the case, that input entities don't have an operator. In that case putting `null` into the `OperatorInput` field in the corresponding input entity would be doable, but not indicating what exactly the `null` stands for (error in entity creation, no operator, ...). Hence, some default fields for the `OperatorInput` might make sense.

ckittl:
Successfully resolved

PR #41, opened 2020-02-13
Addressing #40 - added NO_OPERATOR field in OperatorInput

johanneshiry:
See #40 for details


sensarmad:
If I understand the issue correctly, I will prefer the default value to be NO_OPERATOR_ASSIGNED or DEFINED. Just NO_OPERATOR might also be confused with number(no.) of operators later. Just for clarity's sake.

johanneshiry:
> If I understand the issue correctly, I will prefer the default value to be NO_OPERATOR_ASSIGNED or DEFINED. Just NO_OPERATOR might also be confused with number(no.) of operators later. Just for clarity's sake.

good point - changed the field name accordingly to `NO_OPERATOR_ASSIGNED`

johanneshiry:
!test

PR #50, opened 2020-02-26
Use the correct unit for admittance per length

### LINKED ISSUE #49: Correct the unit for standard admittance per length
Currently it's n(S/km), where µS/km would be desired.

ckittl:
Resolves #49 
Is blocked by ie3-institute/PowerSystemUtils#17


ckittl:
Postpone the adaption of ```ADMITTANCE_PER_LENGTH```-implementation until the next stable version of PowerSystemUtils is released.

PR #60, opened 2020-03-11
Ds/#42 release blocker

### LINKED ISSUE #53: Proper implementation of toString() for Entities
Currently, a lot of the entities (e.g. LoadInput, PvResult, ...) are missing their own implementation of `toString()`. Would be nice if we could provide a proper implementation of them to make them easier to be printed in logs and exceptions.

denstre:
Adds missing toString methods - resolves #53


denstre:
!test

PR #62, opened 2020-03-11
Init InputEntityProcessor

denstre:


ckittl:
Please see #74, #83, #86 before continuing.

PR #83, opened 2020-03-25
Adding processor for graphic inputs

ckittl:
Partly solves #73 (introduces an `GraphicInputProcessor` for all `GraphicInput` entities) and is based on #74.

when this one is merged, we still need processors for remaining `InputEntity`s (`OperatorInput`, `AssetCharacteristicInput`-subclasses, `RandomLoadParameter` and `AssetTypeInput` - _when addressing this, it might be possible to extend the `AssetInputProcessor` to a full `InputEntityProcessor` instead of splitting up logic to new processors_)

Additionally this pull request closes #82.

Issue #73, opened 2020-03-23
Extend Processors to account for all subclasses of UniqueEntity + extend CsvFileSink accordingly 


johanneshiry:
For now, the `CsvFileSink` is only able to process subclasses of `ResultEntity`. The reason is missing implementations of Processors that are able to 'de-serialize' all kinds of `UniqueEntity` to make them usable by the `CsvFileSink`. 

Through creation of more `Processors` we can leverage the full capabilities of the `CsvFileSink` to write all types of `UniqueEntity`.

ckittl:
Duplicate to #61?

PR #86, opened 2020-03-26
Adding processor for other input entities

ckittl:
Partly solves #73 (introduces an `InputEntityProcessor` for all `InputEntity` entities) and is based on #83.

when this one is merged, we still need processors for everything that has to do with input time series.


ckittl:
!test

ckittl:
!test

johanneshiry:
!test

johanneshiry:
!test

docs/uml/main/InputDatamodelConcept.puml
```diff
@@ -36,11 +36,20 @@ abstract Class ConnectorInput {
 }
 ConnectorInput --|> AssetInput
 
-abstract Class AssetCharacteristicsInput {
-+ type: String
+abstract Class AssetCharacteristicsInput<T extends AssetTypeInput> {
```
johanneshiry: Does not match implementation which is `SystemParticipantTypeInput`

docs/uml/main/InputDatamodelConcept.puml
```diff
@@ -36,11 +36,20 @@ abstract Class ConnectorInput {
...
+    - type: T
+    - Characteristic: String
```
johanneshiry:
```suggestion
    - characteristic: String
```

docs/uml/main/InputDatamodelConcept.puml
```diff
@@ -36,11 +36,20 @@ abstract Class ConnectorInput {
...
 }
 AssetCharacteristicsInput --|> InputEntity
 
+Class WecCharacteristicInput {
+}
+WecCharacteristicInput --|> AssetCharacteristicsInput: <<bind>>T::WecTypeInput
```
johanneshiry: Would you mind explaining me what `<<bind>>` means, please?

src/main/java/edu/ie3/datamodel/io/factory/input/NodeInputFactory.java
```diff
@@ -17,10 +17,10 @@
 
 public class NodeInputFactory extends AssetInputEntityFactory<NodeInput, AssetInputEntityData> {
   private static final String V_TARGET = "vtarget";
-  public static final String V_RATED = "vrated";
+  public static final String V_RATED = "vRated";
   private static final String SLACK = "slack";
   private static final String GEO_POSITION = "geoposition";
-  public static final String VOLT_LVL = "voltlvl";
+  public static final String VOLT_LVL = "voltLvl";
```
johanneshiry: Why this change? This someone destroys the initial concept that everything should be considered lowercase in the factories as well as introduces a confusing `toLowerCase()` below, which in fact could be avoided by just writing it lowercase. Especially because the only other place it is used ist the `EntityProcessor` class where all calls are together with an `equalsIgnoreCase()`. Hence I actually see no sense in changing this here as well. Please let me know why this is necessary.

src/main/java/edu/ie3/datamodel/io/factory/input/NodeInputFactory.java
```diff
@@ -43,7 +45,7 @@ protected NodeInput buildModel(
         data.getQuantity(V_TARGET, StandardUnits.TARGET_VOLTAGE_MAGNITUDE);
     final boolean slack = data.getBoolean(SLACK);
     final Point geoPosition = data.getPoint(GEO_POSITION).orElse(null);
-    final VoltageLevel voltLvl = data.getVoltageLvl(VOLT_LVL, V_RATED);
+    final VoltageLevel voltLvl = data.getVoltageLvl(VOLT_LVL.toLowerCase(), V_RATED.toLowerCase());
```
johanneshiry: Together with my last comment could be reverted

docs/uml/test/TestDataHierarchy.puml
```diff
@@ -2,7 +2,9 @@
 Interface GridTestData
```
johanneshiry: :+1:

src/main/java/edu/ie3/datamodel/io/factory/typeinput/SystemParticipantTypeInputFactory.java
```diff
@@ -37,7 +37,7 @@
   private static final String E_CONS = "econs";
 
   // BmTypeInput
-  private static final String LOAD_GRADIENT = "loadgradient";
+  private static final String ACTIVE_POWER_GRADIENT = "activepowergradient";
```
johanneshiry: good renaming! :+1:

src/main/java/edu/ie3/datamodel/io/factory/typeinput/SystemParticipantTypeInputFactory.java
```diff
@@ -50,7 +50,6 @@
 
   // StorageTypeInput
   private static final String P_MAX = "pmax";
-  private static final String CP_RATE = "cprate";
```
johanneshiry: Are these changes addressing already all changes we discussed with @sensarmad? And if yes, are they comprehensive? Or whats the reason for changing them here?

src/main/java/edu/ie3/datamodel/io/factory/typeinput/SystemParticipantTypeInputFactory.java
```diff
@@ -204,7 +210,8 @@ private SystemParticipantTypeInput buildStorageTypeInput(
       double cosPhi) {
     Quantity<Energy> eStorage = data.getQuantity(E_STORAGE, StandardUnits.ENERGY_IN);
     Quantity<Power> pMax = data.getQuantity(P_MAX, StandardUnits.ACTIVE_POWER_IN);
-    Quantity<DimensionlessRate> cprate = data.getQuantity(CP_RATE, StandardUnits.CP_RATE);
+    Quantity<DimensionlessRate> cprate =
```
johanneshiry: If we keep this I would suggest (needs to be adapted in the constructor below (@ line 221) as well 

```suggestion
    Quantity<DimensionlessRate> activePowerGradient =
```

src/main/java/edu/ie3/datamodel/io/processor/EntityProcessor.java
```diff
@@ -53,6 +51,14 @@
   private static final String VOLT_LVL = NodeInputFactory.VOLT_LVL;
   private static final String V_RATED = NodeInputFactory.V_RATED;
 
+  /* Quantities associated to those fields must be treated differently (e.g. input and result), all other quantity /
```
johanneshiry: :+1:

src/main/java/edu/ie3/datamodel/io/processor/EntityProcessor.java
```diff
@@ -337,68 +350,11 @@ protected String processVoltageLevel(VoltageLevel voltageLevel, String fieldName
...
-            quantityValToOptionalString(
-                quantity.asType(EnergyPrice.class).to(StandardUnits.ENERGY_PRICE));
-        break;
-      default:
-        log.error(
-            "Cannot process quantity with value '{}' for field with name {} in model processing!",
-            quantity,
-            fieldName);
-        break;
+    if (specificQuantityFieldNames.contains(fieldName)) {
```
johanneshiry: While I like this approach it neglects that the provided data might not be in `StandardUnits`. From my point of view you can't always ensure that the provided data is in `StandardUnits` because it's NOT mandatory that they have been read by our processors and transformer to standard units accordingly. Hence I would suggest keeping the distinction between quantities that needs to be normalized and the ones that can just be handled via `quantityValtoOptionalString()`

src/main/java/edu/ie3/datamodel/models/input/OperatorInput.java
```diff
@@ -15,16 +15,16 @@
       new OperatorInput(UUID.randomUUID(), "NO_OPERATOR_ASSIGNED");
 
   /** The name of this operator */
```
johanneshiry:
```suggestion
  /** The id (=name) of this operator */
```

src/main/java/edu/ie3/datamodel/models/input/connector/LineInput.java
```diff
@@ -60,7 +61,7 @@ public LineInput(
       Optional<String> olmCharacteristic) {
     super(uuid, operationTime, operator, id, nodeA, nodeB, parallelDevices);
     this.type = type;
-    this.length = length;
+    this.length = length.to(StandardUnits.LINE_LENGTH);
```
johanneshiry: good catch 👍

src/main/java/edu/ie3/datamodel/models/input/system/type/StorageTypeInput.java
```diff
@@ -58,15 +58,15 @@ public StorageTypeInput(
       Quantity<Power> sRated,
       double cosphiRated,
```
johanneshiry: Getter/Setter needs to be addressed accordingly 

```suggestion
      double cosPhiRated,
```

docs/uml/main/InputDatamodelConcept.puml
ckittl: I'm not an UML expert, too. It should underline, that the implementation binds the generic type of `AssetCharacteristicInput` to `WecTypeInput`.

src/main/java/edu/ie3/datamodel/io/factory/input/NodeInputFactory.java
ckittl: I also feel, that it is not too nice. Basically we can break it down, that the factories and processors have to have knowledge about the field names of an entity. This information is unfortunately taken from different sources - the factories utilize the static final fields in the factories themselves, whereas the processors reflect the object's bean properties. As well unfortunately, they both cover inconsistent information (e.g. `geoposition` vs. `geoPosition`).

At this particular place it makes a difference, because the static fields `VOLT_LVL` and `V_RATED` are utilized directly to decompose the `VoltageLevel` object into two fields (cf. the `public` descriptor). Therefore, the headline of a `NodeInput` would be something like `...,slack,geoPosition,voltlvl,vrated,...` instead of `...,slack,geoPosition,voltLvl,vRated,...`.

So, all in all to me this was less worse way of fixing the headline while maintaining the correct function of the factory. But I'm open to other suggestions.

src/main/java/edu/ie3/datamodel/io/factory/typeinput/SystemParticipantTypeInputFactory.java
ckittl: I discussed them all with @sensarmad. It was on the one hand to harmonize naming with the biomass model, which also has a maximum rate of change of power and on the other hand, bringing a more comfortable unit into play for the life time. The initial changes, you have been involved in discussion have already been addressed in a different PR. Do you feel there is something missing?

src/main/java/edu/ie3/datamodel/io/processor/EntityProcessor.java
ckittl: Okay, but then we would have to "standardize" all quantities, we get. I decided not to do it, because the conversion is made within all constructors of our models. So unless you do any reflective magic, all our models will hold the correct unit.

docs/uml/main/InputDatamodelConcept.puml
ckittl: Adapted the implementation instead. ;-)

src/main/java/edu/ie3/datamodel/io/factory/input/NodeInputFactory.java
ckittl: I added #91 to keep an eye on it. 👀

src/main/java/edu/ie3/datamodel/models/input/OperatorInput.java
```diff
@@ -14,17 +14,17 @@
   public static final OperatorInput NO_OPERATOR_ASSIGNED =
       new OperatorInput(UUID.randomUUID(), "NO_OPERATOR_ASSIGNED");
 
-  /** The name of this operator */
-  private final String name;
+  /** The id of this operator */
```
johanneshiry:
```suggestion
  /** The id (=name) of this operator */
```

ckittl:
!test

PR #89, opened 2020-03-27
Adding processor for time series

### LINKED ISSUE #73: Extend Processors to account for all subclasses of UniqueEntity + extend CsvFileSink accordingly

ckittl:
Resolves #73 (introduces an processor for all time series) and is based on #86.

### DIFF
--- a/docs/uml/main/InputDatamodelConcept.puml
+++ b/docs/uml/main/InputDatamodelConcept.puml
@@ -368,39 +368,39 @@ Class PowerValue {
     - p: Quantity<Power> [kW]
     - q: Quantity<Power> [kVA]
 }
-PowerValue --|> Value
+Value <|-- PowerValue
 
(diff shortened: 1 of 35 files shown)


ckittl:
All Codacy issues are addressed. Seems to be some kind of recognition problem.

ckittl:
This branch conceptually drifted in a direction, which is not really fitting into the overall concept. Instead of rolling everything back, I will start a new PR. Therefore, this PR is closed.

PR #92, opened 2020-03-31
Extractor for nested models 

### LINKED ISSUE #78: Extractor for nested Models into their own collections for persistence
Some of our models are nested e.g. a 2 winding transformer holds an instance of its node, while in a database or a file only the primary key is referenced in the corresponding field. For simplification and to avoid missing elements when persisting a model, it would be nice to have an extractor class that ensures that all nested models are extracted in their own collection to be passed to the persistence unit. 

e.g a collection of `Transformer2W` can be "extracted" into collections of `Transformer2W`, `OperatorInput` and `NodeInput`

johanneshiry:
resolves #78

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/extractor/Extractor.java
+++ b/src/main/java/edu/ie3/datamodel/io/extractor/Extractor.java
@@ -2,16 +2,14 @@
  * © 2020. TU Dortmund University,
  * Institute of Energy Systems, Energy Efficiency and Energy Economics,
  * Research group Distribution grid planning and operation
- */
+*/
 package edu.ie3.datamodel.io.extractor;
 
 import edu.ie3.datamodel.exceptions.ExtractorException;
 import edu.ie3.datamodel.models.input.InputEntity;
-
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-
 
 /**
  * //ToDo: Class Description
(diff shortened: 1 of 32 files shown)


src/main/java/edu/ie3/datamodel/io/extractor/Extractor.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * //ToDo: Class Description
+ *
+ * @version 0.1
+ * @since 31.03.20
+ */
+public class Extractor {
```
ckittl, line 22: I would suggest to have a simple static class. So then you could save the instantiation.

```suggestion
public class NestedEntityUtils {

public NestedEntityUtils() {
    throw new IllegalStateException("Do not instantiate static utility class");
}
```

src/main/java/edu/ie3/datamodel/io/extractor/Extractor.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private final List<InputEntity> extractedEntities;
+
+  public Extractor(Nested nestedEntity) throws ExtractorException {
+    this.extractedEntities = extractElements(nestedEntity);
+  }
+
+  private List<InputEntity> extractElements(Nested nestedEntity) throws ExtractorException {
```
ckittl, line 30:
```suggestion
private Set<InputEntity> extractElements(Collection<Nested> nestedEntities) throws ExtractorException {
    return Collections.unmodifiableSet(nestedEntities.stream().map(nested -> {
      if (nestedEntities instanceof Node) {
        return Collections.singletonList(((Node) nestedEntities).getNode());
      } else if (nestedEntities instanceof NodeC) {
        return Collections.singletonList(((NodeC) nestedEntities).getNodeC());
      } else if (nestedEntities instanceof Nodes) {
        return Arrays.asList(((Nodes) nestedEntities).getNodeA(), ((Nodes) nestedEntities).getNodeB());
      } else if (nestedEntities instanceof Type) {
        return Collections.singletonList(((Type) nestedEntities).getType());
      } else {
        return new ArrayList<InputEntity>();
      }
    }).flatMap(List::stream).collect(Collectors.toSet()));
  }
```

This automatically would ensure, that duplicated nodes etc. are only apparent once.

src/main/java/edu/ie3/datamodel/io/extractor/Nested.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This interface should be implemented only by other interfaces that should be used by the {@link
+ * Extractor} It provides the entry point for the extraction method in the {@link Extractor}. If
+ * this interface is implemented by other interfaces one has to take care about, that the
+ * corresponding method {@link Extractor.extractElements()} is extended accordingly.
+ *
+ * @version 0.1
+ * @since 31.03.20
+ */
+public interface Nested {}
```
ckittl, line 17:
```suggestion
public interface NestedEntity {}
```

src/main/java/edu/ie3/datamodel/io/extractor/Node.java
```diff
@@ -0,0 +1,20 @@
...
+import edu.ie3.datamodel.models.input.NodeInput;
+
+/**
+ * Interface that should be implemented by all elements holding a {@link NodeInput} and should be
+ * processable by the {@link Extractor}
+ *
+ * @version 0.1
+ * @since 31.03.20
+ */
+public interface Node extends Nested {
```
ckittl, line 17: What about pimpin the `Nodes` interface and letting go `Node` and `NodeC`? See comments below.

src/main/java/edu/ie3/datamodel/io/extractor/Nodes.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Interface that should be implemented by all elements holding a two {@link NodeInput} elements and
+ * should be processable by the {@link Extractor}. For now, this only holds true by all {@link
+ * edu.ie3.datamodel.models.input.connector.ConnectorInput} and {@link
+ * edu.ie3.datamodel.models.input.connector.TransformerInput}
+ *
+ * @version 0.1
+ * @since 31.03.20
+ */
+public interface Nodes extends Nested {
```
ckittl, line 19:
```suggestion
public interface WithNestedNodes extends Nested {
```
It is not really an Interface describing nodes conceptually, but an entity, that has nodes.

src/main/java/edu/ie3/datamodel/io/extractor/Nodes.java
```diff
@@ -0,0 +1,24 @@
...
+
+  NodeInput getNodeA();
+
+  NodeInput getNodeB();
```
ckittl, line 23:
```suggestion
  Collection<NodeInput> getNodes();
```

PR #93, opened 2020-03-31
Adds a default geoPosition

### LINKED ISSUE #75: Improve NodeInput geoPosition implementation
At the moment, each `NodeInput` holds a non-optional `geoPosition` constructor field. This enforces utilization of `null` entries when no geo information is available. The application of `Optional` is also a not-that-nice way to go as this would cause issues with serialization. Hence, we need a better solution to address cases, when no geo data is available avoiding while avoiding `null` and `Optional`.

Furthermore, `LineInput` still contains an optional for overhead line monitoring. This should be removed asap as well.

denstre:
Partly Resolves #75 
This PR only resolves "NodeInput holds a non-optional geoPosition constructor field"


johanneshiry:
!test

PR #116, opened 2020-04-17
Update README.md

sensarmad:


johanneshiry:
!test

README.md
```diff
@@ -5,4 +5,16 @@
...
+      - Power Grids containing nodes, lines, swithces and transformers
+      - Conventional and renewable generating components such as fixed feed, biomass plants, wind turbines and photovoltaics
+      - Power to heat units e.g. combined heat and power plants and heat pumps
+      - Electrical storages, electric vehicles and charging stations
+      - Thermal units consisting of thermal building and cylindrical storage models
+
+## Supported simulation data:
+      Load profiles, weather data etc.
+      
+Data is represented through CSV files
```
johanneshiry, line 20: That doesn't hold true / is not necessary. *For now* we only provide I/O capabilities for .csv files, but this can easily be extended in the future. Furthermore we *do not* provide data, but only the model. Hence, I would propose removing this line

sensarmad: Modified

Issue #118, opened 2020-04-22
Fix bug in GermanVoltageLevelUtils.parse(...)


johanneshiry:
`CommonVoltageLevel.covers(...)` always throws an exception, even if the correct voltage level could be found. As this method is mainly called by `GermanVoltageLevelUtils.parse(...)` the exception is hidden by the log which is disabled by default in tests. As log level "debug" is disabled in tests, this is not shown in tests but happens there as well. 

To reproduce it just execute `GermanVoltageLevelUtilsTest` and put a breakpoint in `GermanVoltageLevelUtils:122`. Normally, one should **not** end up here for valid data, but one end up here for each element of the test. 

Hence, `CommonVoltageLevel.covers(...)` should be fixed to _not_ throw exceptions for valid data!

denstre:
This section:
`    if (idCovered ^ voltageCovered)
      throw new VoltageLevelException(
          "The provided id \""
              + id
              + "\" and rated voltage \""
              + vRated
              + "\" could possibly meet the voltage level \""
              + this.id
              + "\" ("
              + voltageRange
              + "), but are inconsistent.");
    return idCovered; /* voltage covered is always true, otherwise the exception would have been thrown. */`

has been programmed to throw an exception every time idCovered does not equal voltageCovered.
So every time you iterate through the data and do not find a match, this condition is met and the exception is thrown.

Every time you call 
`boolean voltageCovered = covers(vRated);`. -> vRated(0.4)
You get a different result:
Interval [110.0 kV, 220.0 kV) -> false
Interval [20.0 kV, 30.0 kV) -> false
Interval [10.0 kV, 20.0 kV) - > false
Interval [380.0 kV, 560.0 kV) -> false
Interval [0.0 kV, 10.0 kV) -> true

Same for synonymousIds
[Mittelspannung, ms, ms_30kv, mv, mv_30kV] -> false
[Mittelspannung, ms, ms_10kv, mv, mv_10kV] -> false
[Hochspannung, hs, hv] -> false
[Höchstspannung, ehv, ehv_220kv, hoes, hoes_220kv] -> false
[Höchstspannung, ehv, ehv_380kv, hoes, hoes_380kv] -> false
[Niederspannung, lv, ns] -> true

So every time, this result does not match the given vRated Value
`  public static CommonVoltageLevel parse(String id, ComparableQuantity<ElectricPotential> vRated)
      throws VoltageLevelException {
    return germanVoltageLevels.stream()
[...]

PR #133, opened 2020-05-05
Fixes GermanVoltageLevel.parse bug

denstre:
Adresses #118 
Please read the comment to decide whether you can work on another solution


ckittl:
@johanneshiry 
Please consider my thoughts in [my issue comment](https://github.com/ie3-institute/PowerSystemDataModel/issues/118#issuecomment-624462643). As you raised the "bug", I would like to have your thoughts on this.

src/test/groovy/edu/ie3/datamodel/models/CommonVoltageLevelTest.groovy
```diff
@@ -48,13 +48,4 @@ class CommonVoltageLevelTest extends Specification {
 		expect:
 		!dut.covers("HS", Quantities.getQuantity(110d, KILOVOLT))   // May be false
 	}
-
```
ckittl, line 51: Please alter this test, as not the function has gone, but become a different output. Please test for the return of `false`.

ckittl: The way you resolve this, obviously depends on the outcome of the discussion of the issue.

johanneshiry: Is this still valid @ckittl? I think this can be resolved can't it?

PR #142, opened 2020-05-11
Generify InputContainer interface.allEntitiesAsList()

### LINKED ISSUE #117: Generify InputContainer interface.allEntitiesAsList()
The implementation of `allEnitiesAsList()` currently only allows the return value of strict `UniqueEntity`. It would be nice, if the return value could be something like `T extends UniqueEntity` to allow for returning more specific entities e.g. `SystemParticipantInput` and similar.

A good first way to deal with this could be something like:

```
/** Represents an aggregation of different entities */
public interface InputContainer<T extends UniqueEntity> {

  /** @return unmodifiable List of all entities */
  List<T> allEntitiesAsList();

  /** checks all values using {@link ValidationUtils} */
  void validate();
}
```

denstre:
Resolves #117


johanneshiry:
@denstre is this up for review? If yes pls let us know. I'm unsure bc it is still marked as "Draft".

PR #152, opened 2020-05-27
Remove invalid fields from SwitchResult

### LINKED ISSUE #151: Remove invalid fields in SwitchResult
The current implementation of `SwitchResult` includes the following fields which will always end up to be zero in any case: 

```java 
ComparableQuantity<ElectricCurrent> iAMag,
ComparableQuantity<Angle> iAAng,
ComparableQuantity<ElectricCurrent> iBMag,
ComparableQuantity<Angle> iBAng,
```

The reason for theses fields is, that the superclass of `SwitchResult` is `ConnectorResult` which requires theses fields to be set. 

Hence removing this fields would either require a removal of the superclass `ConnectorResult` from `SwitchResult` OR the introduction of a superclass for connector Results with port information. I think the safer way to go without changing much of the signature would be the second case.

johanneshiry:
Resolves #151 

I finally ended up just changing the superclass of `SwitchResult` to `ResultEntity` because introducing new interfaces or superclasses turned out to cause more harm than good. 

includes:
- new Factory for `SwitchResult` + corresponding tests 
- new test for `ValidationUtils`
- deleted unreachable code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ most of the commonly known power flow calculations
 coordinates or multiple exactly equal coordinates possible
 -  Extended functionality of `GridAndGeoUtils`
 - `CsvFileConnector` is now set up to process either UniqueEntities or only by file name
+- `SwitchResult` superclass changed from `ConnectorResult` to `ResultEntity`
 
 ### Fixed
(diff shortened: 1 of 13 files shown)


johanneshiry:
!test

PR #162, opened 2020-06-05
Introduction of copy builders for immutable grid assets

### LINKED ISSUE #161: Add builder pattern for input models
Due to the immutability approach we're using, copying objects with one or two changed field values becomes more and more complicated as a single method is required for each change. To address this, the [builder pattern ](https://en.wikipedia.org/wiki/Builder_pattern) can be used in an altered form to simplify the copying process (which becomes close to scala case class copy method then).

I would vote for an implementation of at least all input models to avoid unnecessary code blowup in several projects where containers or entities get altered.

johanneshiry:
Resolves #161

### DIFF
--- a/src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
+++ b/src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
@@ -44,4 +44,36 @@ public abstract class UniqueEntity {
   public String toString() {
     return "UniqueEntity{" + "uuid=" + uuid + '}';
   }
+
+  /**
+   * Abstract class for all builder that build child entities of abstract class {@link UniqueEntity}
+   *
+   * @version 0.1
+   * @since 05.06.20
+   */
+  protected abstract static class UniqueEntityCopyBuilder<T extends UniqueEntityBuilder>
+      implements UniqueEntityBuilder {
+
+    private UUID uuid;
+
+    protected UniqueEntityCopyBuilder(UniqueEntity entity) {
+      this.uuid = entity.getUuid();
+    }
+
+    public T uuid(UUID uuid) {
+      this.uuid = uuid;
+      return childInstance();
+    }
+
+    protected UUID getUuid() {
+      return uuid;
+    }
+
+    protected abstract T childInstance();
+  }
+
+  protected interface UniqueEntityBuilder {
+
+    UniqueEntity build();
+  }
 }
--- a/src/main/java/edu/ie3/datamodel/models/input/connector/Transformer3WInput.java
+++ b/src/main/java/edu/ie3/datamodel/models/input/connector/Transformer3WInput.java
@@ -269,6 +269,7 @@ public class Transformer3WInput extends TransformerInput implements HasType {
 
     private Transformer3WTypeInput type;
     private NodeInput nodeC;
+    private boolean intSlack;
     private final NodeInput internalNode;
 
     private Transformer3WInputCopyBuilder(Transformer3WInput entity) {
@@ -276,6 +277,7 @@ public class Transformer3WInput extends TransformerInput implements HasType {
       this.type = entity.getType();
       this.nodeC = entity.getNodeC();
       this.internalNode = entity.getNodeInternal();
+      this.intSlack = entity.getNodeInternal().isSlack();
     }
 
     @Override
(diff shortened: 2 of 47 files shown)


src/main/java/edu/ie3/datamodel/models/input/connector/Transformer3WInput.java
```diff
@@ -209,4 +255,66 @@ public String toString() {
...
+    private NodeInput nodeC;
+    private boolean intSlack;
+    private final NodeInput internalNode;
+
+    private Transformer3WInputCopyBuilder(Transformer3WInput entity) {
+      super(entity);
+      this.type = entity.getType();
+      this.nodeC = entity.getNodeC();
+      this.internalNode = entity.getNodeInternal();
+      this.intSlack = entity.getNodeInternal().isSlack();
```
ckittl, line 280: This could be somehow misleading in terms of data type `int`.

src/main/java/edu/ie3/datamodel/utils/ContainerUtils.java
```diff
@@ -524,8 +525,10 @@ public static SubGridContainer withTrafoNodeAsSlack(final SubGridContainer subGr
         subGridContainer.getRawGrid().getTransformer2Ws().stream()
             .map(
                 oldTrafo2w ->
-                    copyTrafo2WUpdateNodeA(
-                        oldTrafo2w, oldToNewTrafo2WANodes.get(oldTrafo2w.getNodeA())))
+                    oldTrafo2w
+                        .copy()
+                        .nodeA(oldToNewTrafo2WANodes.get(oldTrafo2w.getNodeA()))
```
ckittl, line 530: A lot of "old" here. I would suggest to get the new node and assign it to a variable to have it a bit more explicitly.

johanneshiry:
Thanks for your remarks. Addressed them + added copy builder for all system participants + according tests

johanneshiry:
!test

johanneshiry:
!test

src/main/java/edu/ie3/datamodel/models/input/connector/Transformer3WInput.java
```diff
@@ -186,6 +228,10 @@ public NodeInput getNodeInternal() {
     return nodeInternal;
   }
 
+  public Transformer3WInputCopyBuilder copy() {
```
ckittl, line 231:
```suggestion
  public Transformer3WInputCopyBuilder getCopyBuilder() {
```

From my point of view this thing is not to vital. I was just wondering, if you would expect return type `Transformer3WInput` and getting a clean copy, instead of a builder. Especially, as no doc is given, what this method is meant to do.

Additionally, we should spend some thought on, if the `copy`-method shouldn't be part of abstract class `AssetInput`?

src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
```diff
@@ -44,4 +44,36 @@ public int hashCode() {
...
+    public T uuid(UUID uuid) {
+      this.uuid = uuid;
+      return childInstance();
+    }
+
+    protected UUID getUuid() {
+      return uuid;
+    }
+
+    protected abstract T childInstance();
```
ckittl, line 72: Please add some Documentation on how this pattern / concept works and what the different methods do. Please also consider writing some docs for ReadTheDocs on how to properly use the concept.

src/test/groovy/edu/ie3/datamodel/models/input/system/StorageInputTest.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+
+class StorageInputTest extends Specification {
+
+	def "A StorageInput copy method should work as expected"() {
+		given:
+		def storageInput = SystemParticipantTestData.storageInput
+
+		when:
+		def alteredUnit = storageInput.copy().type(SystemParticipantTestData.storageTypeInput).build()
```
ckittl, line 19: `#behaviour()` is not tested.

johanneshiry: by purpose because it's deprecated and marked accordingly. There's no sense in writing tests for features that are not supported anymore.

src/main/java/edu/ie3/datamodel/models/input/connector/Transformer3WInput.java
johanneshiry: I would prefer stay with `copy()` as the return type is clear by type safety. Furthermore having `copy().build()` for an unmodified copy is IMHO very expressive.

PR #166, opened 2020-06-08
Adds Regex to modify JSON Stings to match csv spec.

### LINKED ISSUE #141: Add quotation marks to all special fields when writing .csv files
Currently, we're writing out JSON- and JSON-alike-Strings without any quotation marks into .csv-files. While this is sufficient for I/O operations in general, it makes manual viewing and manipulating of existing data very unhandy because if the .csv separator is a comma and the json-string contains commas, most (all?) .csv-table-reader cannot display the .csv-file correctly. 

I _assume_ that adding quotation marks at the beginning and the end of the strings _could_ solve the problem. 

When addressing this, we have to keep the following in mind:

- this might cause the need to adapt I/O operations (especially reading .csv-files) 
- this might cause the need to adapt several I/O tests to ensure that the new handling is working

IMHO the following Strings are affected: 

- Point
- LineString
- Characteristics 

Please add missing Strings that are also affected, if any. 

**Edit:**
After investigating further, it also seems that double quotes are not sufficient for nodeGraphicInput and lineGraphicInput files, but single quotes work. We might consider using single quotes for all files then. 

Solution:
According to the csv specification, single quotes (') are not valid, so that double quotes (") must be used.
To escape double quotes ("), as they would otherwise cause problems in the JSON string, double double quotes ("") are specified.

Example:
{"type":"LineString","coordinates":[[7.411111,51.492528],[7.414116,51.484136]],"crs":{"type":"name","properties":{"name":"EPSG:4326"}}}
converts to:
"{""type"":""LineString"",""coordinates"":[[7.411111,51.492528],[7.414116,51.484136]],""crs"":{""type"":""name"",""properties"":{""name"":""EPSG:4326""}}}"

TODOS:

- [x] Adjust FileSink to double quote all csv data objects
- [x] Determine JSON String and double quote double quotes
- [ ] Adjust Factory to remove all double quotes

denstre:
TODO: (Reminder for myself)

Resolves #141 
- [x] Adapt Tests

- [x] Adapt Factory (convert double double quotes ("") to double quotes("))


src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -21,7 +21,7 @@
   /** Information on the shape of the file */
   private final CsvFileDefinition fileDefinition;
   /** True, if every entry should be quoted */
-  private final boolean quoted;
+  private final boolean quoted = true;
```
denstre, line 24: With double double quoting JSON string, quoted should always remain true.
How should we handle this?

src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -47,7 +47,7 @@ public BufferedCsvWriter(
                 baseFolder + File.separator + fileDefinition.getFilePath(), append),
             StandardCharsets.UTF_8));
     this.fileDefinition = fileDefinition;
-    this.quoted = quoted;
+    this.quoted = true;
```
ckittl, line 50: Please remove `quoted`. The constructor parameter as well as the attribute itself, as we intend to identify per String, if a quoting is necessary or not.

src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -106,6 +106,11 @@ private void writeFileHeader(String[] headLineElements) throws IOException {
   private void writeOneLine(String[] entries) throws IOException {
     for (int i = 0; i < entries.length; i++) {
       String attribute = entries[i];
+      if (attribute.matches("^\"\\{(?:.*)\\}\"$")) {
+        attribute = attribute.replaceAll("\"", "\"\"");
+        attribute = attribute.substring(1, attribute.length() - 1);
+      }
```
ckittl, line 112: Please make this a distinct method, that later can be moved to `StringUtils` in PowerSystemUtils. This method has the following tasks:

1. Identify, if the String has to be quoted: It contains the csv separator character, the character " or any other special character (\n, ...)
2. Replace all single " by ""
3. Add quotation marks at the beginning and the end of the string (you can reuse the Regex in `StringUtils#quote(String)`)

Additionally, please code an equivalent method, that does revert that steps.

src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -106,6 +106,11 @@ private void writeFileHeader(String[] headLineElements) throws IOException {
...
+      System.out.println(attribute);
```
ckittl, line 113: Please don't forget to remove this.

src/main/java/edu/ie3/datamodel/io/factory/EntityData.java
```diff
@@ -80,7 +80,7 @@ public String getField(String field) {
     if (!fieldsToAttributes.containsKey(field))
       throw new FactoryException(String.format("Field \"%s\" not found in EntityData", field));
 
-    return fieldsToAttributes.get(field);
+    return fieldsToAttributes.get(field).replaceAll("\"\"", "\"");
```
ckittl, line 83: I think this is not comprehensive, as e.g. the surrounding quotes are not treated (see comment in `BufferedCsvWriter`).

src/test/groovy/edu/ie3/datamodel/io/factory/input/LineInputFactoryTest.groovy
```diff
@@ -174,4 +174,42 @@ class LineInputFactoryTest extends Specification implements FactoryTestHelper {
 		"{ \"type\": \"LineString\", \"coordinates\": [[7.411111, 51.49228],[7.411111, 51.49228],[7.411111, 51.49228],[7.411111, 51.49228]]}" | _
 		"{ \"type\": \"LineString\", \"coordinates\": [[7.411111, 51.49228],[7.411111, 51.49228],[7.311111, 51.49228],[7.511111, 51.49228]]}" | _
 	}
+
+	def "A LineInputFactory should parse a valid LineInput with different double double quoted geoPosition strings correctly"() {
```
ckittl, line 178: When the above comments are addressed, please make this a test, that works on the String manipulation method and not on the factory, as the factory will only be calling the aforementioned method.

src/test/groovy/edu/ie3/datamodel/io/factory/input/NodeInputFactoryTest.groovy
```diff
@@ -64,4 +64,43 @@ class NodeInputFactoryTest extends Specification implements FactoryTestHelper {
 			assert subnet == Integer.parseInt(parameter["subnet"])
 		}
 	}
+
+	def "A NodeInputFactory should parse a valid NodeInput with double double quotes correctly"() {
```
ckittl, line 68: Same here.

src/test/groovy/edu/ie3/test/helper/FactoryTestHelper.groovy
```diff
@@ -18,6 +18,7 @@ trait FactoryTestHelper {
 	}
 
 	static getGeometry(String value) {
+		value = value.replaceAll("\"\"", "\"")
```
ckittl, line 21: This should use the "unwrap" method as well.

ckittl:
Additionally: Please move the quotation handling from `BufferedCsvWriter` to `CsvFileSink`. The sink suits better, as it could use different other csv writers via it's connector. Then the quotation handling would have to be implemented in every future implementation of the writer.

denstre:
!test

denstre:
!test

denstre:
!test

johanneshiry:
!test

johanneshiry:
Won't compile until https://github.com/ie3-institute/PowerSystemUtils/pull/33 is merged and deployed

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -141,7 +142,7 @@ public CsvDataSource(String csvSep, String folderPath, FileNamingStrategy fileNa
             csvRow
                 .replaceAll(charInputRegex, charReplacement)
                 .replaceAll(geoJsonRegex, geoReplacement)
-                .replaceAll("\"", "")
+                .replaceAll("\"*", "") // remove all quotes from
```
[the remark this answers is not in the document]
johanneshiry: Agree. The workaround with json and characteristics was built exactly for that reason (invalid .csv input data) a few weeks ago. 

With the newly introduced, updated and valid .csv string version, we can proceed as you propose.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -78,15 +78,33 @@ public CsvDataSource(String csvSep, String folderPath, FileNamingStrategy fileNa
     TreeMap<String, String> insensitiveFieldsToAttributes =
         new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
 
-    final String[] fieldVals = fieldVals(csvSep, csvRow);
+    // when replacing deprecated workaround add final modifier before parseCsvRow as well as remove
```
ckittl, line 81: Pls mark this as TODO, then we directly have some support by CI and IDE.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -78,15 +78,33 @@ public CsvDataSource(String csvSep, String folderPath, FileNamingStrategy fileNa
...
+      // try to parse old structure
+      fieldVals = oldFieldVals(csvSep, csvRow);
+      // if this works log a warning to inform the user that this will not work much longer,
+      // otherwise
+      // parsing will fail regularly as expected below
+      if (fieldVals.length == headline.length)
+        log.warn(
+            "You are using an outdated version of the data "
+                + "model with invalid formatted csv rows. This is okay for now, but please updated your files, as the "
+                + "support for the old model will be removed soon.");
```
ckittl, line 96: What is if the `fieldVals` still don't have the same length? Is this covered well? Are lines 102...107 failing in any case? I think the case, where the entries are shorter as the head line is not covered well, isn't it?

ckittl: Understood now, what happens.

PR #171, opened 2020-06-22
Removes behaviour from StorageInput

### LINKED ISSUE #136: Remove `behaviour` from `StorageInput`
... as this is only of interest for a smaller portion of simulators. Therefore, they should be responsible of the correct parameterization.

denstre:
Resolves #136 

SIMONA requires refactoring before merge


src/main/java/edu/ie3/datamodel/io/factory/input/participant/StorageInputFactory.java
```diff
@@ -16,15 +16,14 @@
...
     extends SystemParticipantInputEntityFactory<
         StorageInput, SystemParticipantTypedEntityData<StorageTypeInput>> {
-  private static final String BEHAVIOUR = "behaviour";
 
   public StorageInputFactory() {
     super(StorageInput.class);
   }
 
   @Override
   protected String[] getAdditionalFields() {
```
denstre, line 25: Is this method still necessary

PR #168, opened 2020-06-10
Finalizing pull requests

ckittl:
Resolves #158

PR #173, opened 2020-06-24
Fix integration test task

### LINKED ISSUE #144: Review the gradle test tasks
The integrationTests tasks does not seem to work atm. Maybe it has something to do with using spock instead of junit as it works(/-ed?) on DWDTools where the integration tests are junit tests.

ckittl:
This pull request closes #144 .

Additionally, it fixes a path mistake on generating png files from plantuml files. Please don't be afraid of the high amount of changes. It comes from re-generating the JavaDoc. The current state in master has been generated with Java 11, which generates HTML 5 doc, whereas Java 8 generates HTML 4 docs. As this project relies on Java 8 I would opt for having the JavaDoc from this source.


ckittl:
Please also note the changes in test task definitions introduced in #168.

Issue #119, opened 2020-04-23
Change from tec.uom.se to tech.units.indriya (Units of Measurement)


denstre:
tec.uom.se is outdated and won't receive any updates or maintenances.
tech.units.indriya is the successor of tec.uom.se

**Warning: Working on this issue and finally merging it will be break compatibility to previous versions and hence should only be done with a new major release!**

PR #178, opened 2020-07-13
Introduces tech.units.indriya

denstre:
Refactors to new Units of Measurement

#119


ckittl:
Please close #190 before addressing these changes.

mia-krause:
Waiting for ie3-institute/PowerSystemUtils#45 to be merged

mia-krause:
@johanneshiry  @ckittl are you fine with ignoring the sonarqube critic? Then I'll edit it in sonarqube itself.

src/main/java/edu/ie3/datamodel/models/input/NodeInput.java
```diff
@@ -121,15 +122,17 @@ public NodeInputCopyBuilder copy() {
...
-  @Override
   public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     if (!super.equals(o)) return false;
     NodeInput nodeInput = (NodeInput) o;
+    if (!QuantityUtil.quantityIsEmpty(vTarget)) {
+      if (QuantityUtil.quantityIsEmpty(nodeInput.vTarget)) return false;
+      if (!vTarget.isEquivalentTo(nodeInput.vTarget)) return false;
+    } else if (!QuantityUtil.quantityIsEmpty(nodeInput.vTarget)) return false;
```
ckittl, line 133: I would like you to refactor this piece of code, because it has multiple implications:

Please move this code to `QuantityUtils` in PowerSystemUtils and give it a meaningful name (e.g. `isTheSameConsideringEmpty` or `deprecatedEqualsConsideringEmpty`). There, your can also write tests for this method. This would also lower the number of untested lines, here. Please use the javadoc there, to unambiguously state, why the implementation is in the way it is. For me, it is highly unclear - or better to say misleading - what `isEquivalentTo` actually does (to me equivalence is 1 km == 1,000 m). Furthermore all those `if`-branches impose a high cognitive complexity, imposing code smells as well.

Sorry for changing out minds, but it harms more, than the solution we have had before. We just didn't saw that back then.

Please replace the single constructs here and in other classes with the aforementioned implementation.

src/test/groovy/edu/ie3/datamodel/models/input/NodeInputTest.groovy
```diff
@@ -32,4 +35,25 @@ class NodeInputTest extends Specification {
...
+
+		then:
+		nodeA.equals(nodeB) == expectedResult
+
+		where:
+		nodeA                                                                            | nodeB                                                                                            || expectedResult
+		GridTestData.nodeB                                                               | GridTestData.nodeB                                                                               || true
+		GridTestData.nodeB                                                               | GridTestData.nodeB.copy().build()                                                                || true
+		GridTestData.nodeB                                                               | GridTestData.nodeB.copy().vTarget(Quantities.getQuantity(1, PowerSystemUnits.PU)).build()        || true
+		GridTestData.nodeB                                                               | GridTestData.nodeB.copy().vTarget(Quantities.getQuantity(100, PowerSystemUnits.PERCENT)).build() || true
```
ckittl, line 53: Did you validate it against the old implementation? I did not, but as far as I understood the old implementation, diverging Units would lead to two objects not being equal. Please double check this.

src/test/groovy/edu/ie3/datamodel/utils/GridAndGeoUtilsTest.groovy
```diff
@@ -78,6 +78,6 @@ class GridAndGeoUtilsTest extends Specification {
 		def nodeB = GridTestData.nodeB
 
 		expect:
-		GridAndGeoUtils.distanceBetweenNodes(nodeA, nodeB) == Quantities.getQuantity(0.9135678707610981, PowerSystemUnits.KILOMETRE)
+		GridAndGeoUtils.distanceBetweenNodes(nodeA, nodeB) == Quantities.getQuantity(0.91356787076109815268517, PowerSystemUnits.KILOMETRE)
```
ckittl, line 81: I would like to have this refactored using `QuantityUtils.considerablyEqual()`

ckittl (changes requested):

ckittl:
> @johanneshiry @ckittl are you fine with ignoring the sonarqube critic? Then I'll edit it in sonarqube itself.

I think, with the requested changes, the quality gate should pass.

src/main/java/edu/ie3/datamodel/models/input/NodeInput.java
mia-krause: To clarify: isEquivalentTo works exactly like you thought, (1 km).isEquivalentTo(1000m) == true. And also, more importantly considering the issues we had in the past, (1 km).isEquivalentTo(1.0 km).
So, just to have it clear at this point, should the equals methods of our DataModels return true or false in your example case? And if false, should they return true in an Int/Double case?

ckittl:
Replaced by #206

PR #196, opened 2020-09-01
jh/#195 add charging station model

### LINKED ISSUE #195: Add support for ev charging stations
At the moment we have no valid implementation (dummy only) support for electric vehicle charging stations as system participants. As per requirement of some research projects and because of the rising share of electric vehicles and corresponding charging infrastructure that is expected within the next few years, supporting charging infrastructure in the data model is something that we need to address.

For now, this mainly means a) developing a simple model to represent charging infrastructure within the power system data model and b) replace the existing dummy implementations with the developed model.

johanneshiry:
Resolves #195


build.gradle
```diff
@@ -52,7 +52,7 @@ repositories {
 
 dependencies {
 	//	 ie³ power system utils
-	compile 'com.github.ie3-institute:PowerSystemUtils:1.3.2-SNAPSHOT'
+	compile 'com.github.ie3-institute:PowerSystemUtils:1.3.2'
```
ckittl, line 55: Is this really the final PR before changing the branch and workflow structure?

docs/readthedocs/models/input/participant/evcs.rst
```diff
@@ -1,5 +1,123 @@
...
++------------------------+---------+--------------------------------------------------------------------------------+
+
+.. _evcs_point_types:
+
+Charging Point Types
+^^^^^^^^^^^^^^^^^^^^
+Available Standard Types
+""""""""""""""""""""""""
+To simplify the application of electric vehicle charging stations, some common standard types are available out-of-the-box.
+They can either by used code vise or directly from database or file input by referencing their id or one of their
```
ckittl, line 66:
```suggestion
They can either by used code wise or directly from database or file input by referencing their id or one of their
```
❓

docs/readthedocs/models/input/participant/evcs.rst
```diff
@@ -1,5 +1,123 @@
...
+
+
+Custom Types
+""""""""""""
+While the provided standard types should be suitable for most scenarios, providing an individual type for a specific
+scenario might be necessary. To do so, a custom type can be provided instead of a common id. This custom type is tested
+against the following regex :code:`(\w+\d*)\s*\(\s*(\d+\.?\d+)\s*\|\s*(AC|DC)\s*\)`, or more generally, the custom
+type string has to be in the following syntax::
+
+    <Name>(<Apparent Power in kVA>|<AC|DC>) e.g. FastCharger(50|DC) or SlowCharger(2.5|AC)
```
ckittl, line 105: We should consider using similar typing of characteristics, don't we? In most other cases we use something, that is JSON-like. This implies no favour for either solution, but if you agree, you might hand in an issue proposing to harmonise notations.

docs/readthedocs/models/input/participant/evcs.rst
```diff
@@ -1,5 +1,123 @@
...
+
+Please note, that in accordance with :code:`edu.ie3.datamodel.models.StandardUnits` the apparent power is expected to
```
ckittl, line 107: I would suggest to have an different version of the regex:
`(\w+\d*)\ ?\(\ ?(\d+(?:\.\d+)?\ ?[kMG]?W)\ ?\|\ ?(AC|DC)\ ?\)`

- `\s+` would also allow tabs and stuff, additionally in any amount. `\ ?` would only allow one blank at most
- Explicitly demand for a power string (e.g. '50 kW', '5000 W', '0.5GW', that can be parsed to quantity directly.

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointType.java
```diff
@@ -0,0 +1,102 @@
...
+            })
+        .orElseGet(
+            () ->
+                id
+                    + "("
+                    + sRated.to(PowerSystemUnits.KILOVOLTAMPERE).getValue().doubleValue()
+                    + "|"
+                    + electricCurrentType
+                    + ")");
+  }
```
ckittl, line 101: Could you shed some light on, why this method is that complex? This implies, that there might be a type, that has the same id as a a common type and also the same properties (so a duplicate of the common ones). Then the common one is acquired and it's id is given back. In that case only the id is the result of the `toString`-Method, while in any other case the String looks a bit more detailed?!

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
```diff
@@ -0,0 +1,246 @@
...
+          Quantities.getQuantity(150, StandardUnits.S_RATED),
+          DC,
+          new HashSet<>(Arrays.asList("tesla2", "teslav2", "supercharger2")));
+
+  public static final ChargingPointType TeslaSuperChargerV3 =
+      new ChargingPointType(
+          "TeslaSuperChargerV3",
+          Quantities.getQuantity(250, StandardUnits.S_RATED),
+          DC,
+          new HashSet<>(Arrays.asList("tesla3", "teslav3", "supercharger3")));
```
ckittl, line 122: Is there a reason not to implement them as an enum?

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
```diff
@@ -0,0 +1,246 @@
...
+
+  protected static final Map<String, ChargingPointType> commonChargingPointTypes =
```
ckittl, line 124: Pls add JavaDoc. I think, it just concats all available ids and synonymous ids and maps them to the actual charging point types?

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
```diff
@@ -0,0 +1,246 @@
...
+   * @return instance of a {@link ChargingPointType}
+   * @throws ChargingPointTypeException if the string is neither a valid custom charging point
+   *     string nor a valid common standard charging point id
+   */
+  public static ChargingPointType parse(String parsableString) throws ChargingPointTypeException {
+
+    ChargingPointType res;
+
+    // valid regex for either custom or pre-defined types
+    String validCustomRegex = "(\\w+\\d*)\\s*\\(\\s*(\\d+\\.?\\d+)\\s*\\|\\s*(AC|DC)\\s*\\)";
```
ckittl, line 169: See comment / suggestion in documentation.

ckittl (changes requested):

build.gradle
ckittl: 😂 What a dumb comment. Forget about it. 🤦

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
johanneshiry: There are two tbh: 

1. It is based on the same structure as `VoltageLevel` - I only adapted the concept
2. Having this implemented as an enum would IMHO prevent the option to have a custom charging point type provided

docs/readthedocs/models/input/participant/evcs.rst
johanneshiry: Agree. 
Choosing this approach over the existing ones (either having it's own type file, a JSON- or a JSON-like-String) has been motivated by the number of attributes (id, apparent power and current) which is with 3 (actual 2) very much nothing. So having it as JSON-/JSON-like string would cause a way longer string which IMHO highly decrease readability. On the other hand, having a type file with only 3 columns but the huge overhead this creates when reading and writing types I also thought this might be overkill for this small piece of information. 

tl;dr

The main reasons for choosing this approach over JSON-/JSON-like or type file are:

- better readability
- no overhead for code adaptions (file hierarchy, file types, sinks, sources, ...)
- not breaking old data models bc if missing charging station support 


I'm not sure, if harmonizing syntax makes sense at this point though. Characteristics (and I assume you refer to WEC characteristics or similar) are more or less time series with way more and different information than the type information. I'm open to discuss this and hand in an issue for harmonizing but maybe you could elaborate a little bit more why harmonizing would make sense here?!

ckittl: ack

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
ckittl: ack

docs/readthedocs/models/input/participant/evcs.rst
johanneshiry: While I really like this idea and its verbosity I would prefer not implementing this approach for now as this requires several things to be changed incl. 

- ensuring that there is always a white space between the value and the unit (`50 kW` can be parsed by the Unit API, `50kW` throws an exception)
- requires casting of the resulting value using `asType(Power.class)`
- still requires conversion to `StandardUnits.S_RATED` using `to(StandardUnits.S_RATED)`
- several test adaptions

All these changes for a small verbosity improvement I would prefer keeping the existing solution and if there will be a point where having a more explicit mapping is required we could address this.

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java
johanneshiry: see [answer](https://github.com/ie3-institute/PowerSystemDataModel/pull/196#discussion_r504565700)

src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointType.java
johanneshiry: The idea behind this implementation was, that whenever this type is a common type, the only the easy to read, common id is given back instead of the full id with sRated and electric current type. In contrast, when this is a custom type, we actually need all information, because custom types have custom attributes (sRated, current type).

ckittl (dismissed):

ckittl (approved):

PR #206, opened 2020-09-11
jh/#119-switch-to-indriya-quantities

### LINKED ISSUE #119: Change from tec.uom.se to tech.units.indriya (Units of Measurement)

johanneshiry:
Replacement for #178 

Actually work on this can start when PowerSystemDataModel is on 1.4-SNAPSHOT and the corresponding PR [#46](https://github.com/ie3-institute/PowerSystemUtils/pull/46) is merged and deployed. Including the breaking dependency will then reveal all places that require actions to be taken and hence is easier and safer to address as if the work starts as long as the old, deprecated stuff is still available. 

Resolves #119


ckittl:
Pls delete branch `ds/#119-change-units-of-measurement` as soon as you know, that you don't need the work anymore.

ckittl:
@mia-krause  @johanneshiry I found some stuff, that might be in (at least slight) conjunction with ie3-institute/powersystemutils#49.

I experienced failing tests, e.g. `CsvRawGridSourceTest`: When a line type model is set up from file, the relative susceptance is created with e.g. `Quantities.getQuantity(0.00322, MICRO_SIEMENS_PER_KILOMETRE)`, whereas the expected line type is created with `Quantities.getQuantity(0.00000000322, SIEMENS_PER_KILOMETRE)` (the constructor takes care of converting it to µS/km). However, as the new `equals` method of indriya lib does check for real equality, both models are not equal.

This imposes some questions regarding all classes, that do have quantities as attributes:

- [ ] We actually would like the models `equals` check the quantities to be equivalent, don't we?
- [ ] Would we like to check this with or without tolerance (I would opt for "without")?

johanneshiry:
@mia-krause I had to modify the influxDb implementation as the `TimeBasedWeatherValueFactory` didn't provided a uuid to the resulting `TimeBasedValue` which is actually not what we want to have. Please carefully check the changed implementation as well as commit [a6e4180cf5e034f32c91ef75d4b6ea1abadca60a](https://github.com/ie3-institute/PowerSystemDataModel/pull/206/commits/a6e4180cf5e034f32c91ef75d4b6ea1abadca60a)

ckittl:
Is this ready for review, yet?

docs/readthedocs/models/models.rst
```diff
@@ -48,6 +48,31 @@ Harmonized Units System
...
+      the entities :code:`equals()` method returns false. This behavior is in sync with the equals implementation
+      of the indriya library. Hence, you should ensure that your code always pass in the same kind of a quantity instance
+      with the same underlying number format and type.
+    - for now, there is no default way to compare entities in a 'number equality' way provided. E.g. a line with a length
+      of 1km compared to a line with a length of 1000m is actually of the same length, but calling :code:`LineA.equals(LineB)`
+      would return :code:`false` as the equality check does NOT convert units. If you want to compare two entity instances
+      based on their equality you have (for now) check for each quantity manually using their :code:`isEquivalentTo()`
+      method. If you think you would benefit from a standard method that allows entity equivalence check, please consider
+      handing in an issue `here <https://github.com/ie3-institute/PowerSystemDataModel/issues>`_.
+
```
ckittl, line 75: Is it advisory to spend a word on comparison with tolerance?

src/main/java/edu/ie3/datamodel/io/processor/Processor.java
```diff
@@ -328,12 +329,21 @@ protected String processVoltageLevel(VoltageLevel voltageLevel, String fieldName
...
     } else {
-      return quantityValToOptionalString(quantity);
+      optQuant = quantityValToOptionalString(quantity);
     }
+    return optQuant.orElseThrow(
+        () ->
+            new EntityProcessorException(
+                "Unable to process quantity value for attribute '"
+                    + fieldName
+                    + "' in result entity "
```
ckittl, line 344:
```suggestion
                    + "' in entity "
```

src/test/groovy/edu/ie3/datamodel/models/value/IrradiationValueTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+		new IrradiationValue(Quantities.getQuantity(10d, StandardUnits.IRRADIATION), null)                   | new IrradiationValue(null, null)                                                     || false
+		null                                                                                                 | new IrradiationValue(null, null)                                                     || false
+		null                                                                                                 | null                                                                                 || true
+		new IrradiationValue(Quantities.getQuantity(10.23d, StandardUnits.IRRADIATION), null)                | new IrradiationValue(Quantities.getQuantity(10.23, StandardUnits.IRRADIATION), null) || false
+		new IrradiationValue(Quantities.getQuantity(10230, PowerSystemUnits.WATTHOUR_PER_SQUAREMETRE), null) | new IrradiationValue(Quantities.getQuantity(10.23, StandardUnits.IRRADIATION), null) || false
+	}
+}
```
ckittl, line 30: Do we need to fill up all other (yet missing) tests as well?

ckittl (changes requested):
I haven't seen any places, that I'm afraid of breaking something with regard to time series.

docs/readthedocs/models/models.rst
```diff
@@ -48,6 +48,31 @@ Harmonized Units System
...
+      Using :code:`Objects.equals(<UnitInstanceA>, <UnitInstanceB>)` is necessary especially for time series data.
+      As in contrast to all other places, quantity time series from real world data sometimes are not complete and
+      hence contain missing values. To represent missing values this is the only place where the usage of :code:`null`
+      is a valid choice and hence needs to be treated accordingly. Please remember that his is only allowed in very few
+      places and you should try to avoid using :code:`null` for quantities or any other constructor parameter whenever possible!
+    - equality is given if, and only if, the quantities value object and unit are exactly equal. Value objects can become
+      e.g. :code:`BigDecimal` or :code:`Number` instances. It is important, that the object type is also the same, otherwise
+      the entities :code:`equals()` method returns false. This behavior is in sync with the equals implementation
+      of the indriya library. Hence, you should ensure that your code always pass in the same kind of a quantity instance
+      with the same underlying number format and type.
```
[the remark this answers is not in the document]
johanneshiry: Can you provide an example sentence? I'm actually unsure what exactly you mean by "conversion method"

src/test/groovy/edu/ie3/datamodel/models/value/IrradiationValueTest.groovy
johanneshiry: Depends - I would say no, bc I think it's not necessary at all. But if you and @mia-krause think it is necessary/make sense I could add them. Just let me know pls.

docs/readthedocs/models/models.rst
johanneshiry: yep - will add

src/test/groovy/edu/ie3/datamodel/models/value/IrradiationValueTest.groovy
mia-krause: If you want to be really strict about it, they definitely would enforce the nullable equals on weather values. However, I don't think (or hope..) that we will change the equals methods a lot in the future and it also wouldn't enforce the pattern on new entities. So I personally don't think that we need the other tests.

ckittl (approved):

ckittl (approved):

PR #208, opened 2020-09-14
Provide utility routines to check and create a recommended folder structure

### LINKED ISSUE #202: Define and implement a recommended folder structure for input data
- [x] Define a thorough folder structure for input data
- [x] Implement checks and creation routine
- [x] Document in RTD
- [x] Implement routines to extract / archive this structure in `*.tar.gz`-files
- [x] Combine `DefaultFileHierarchy` with `NamingStrategy`

ckittl:
Closes #202


ckittl:
@johanneshiry 
I would like you to have a first look on the concept, before I approach the next steps. Those would be

- incorporation of the result file directory structure
- inclusion of the new hierarchic naming strategy into the broader concept
- maybe some additional parameters to toggle between flat output / input and hierarchic

src/main/java/edu/ie3/datamodel/io/csv/DefaultInputHierarchy.java
```diff
@@ -0,0 +1,271 @@
...
+import org.apache.commons.io.FilenameUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Default directory hierarchy for input models */
+public class DefaultInputHierarchy implements FileHierarchy {
+  private static final Logger logger = LoggerFactory.getLogger(DefaultInputHierarchy.class);
+
+  /** Use the unix file separator here. */
+  protected static final String FILE_SEPARATOR = "/";
```
johanneshiry, line 48: any reason why this is used here instead of `File.separator();`? Does this work within all OS?

src/main/java/edu/ie3/datamodel/io/csv/FileHierarchy.java
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Determines the correct subdirectory (w.r.t. an arbitrary base directory) for a certain given
+   * class using the Unix file separator for delimiting between directories and files.
+   *
+   * @param cls Class to define the sub directory for
+   * @return An Option to the regarding sub directory as a string
+   */
+  default Optional<String> getSubDirectory(Class<? extends UniqueEntity> cls) {
+    return getSubDirectory(cls, "/");
```
johanneshiry, line 34: Again hard coded file separator - I would expected issues here? Why not using `File.separator();`

src/main/java/edu/ie3/datamodel/io/csv/HierarchicFileNamingStrategy.java
```diff
@@ -0,0 +1,103 @@
...
+  public Optional<String> getFileName(Class<? extends UniqueEntity> cls) {
+    /* Get the file name */
+    Optional<String> maybeFilename = super.getFileName(cls);
+    String fileName;
+    if (!maybeFilename.isPresent()) {
+      logger.debug("Cannot determine file name for class '{}'.", cls);
+      return Optional.empty();
+    } else {
+      /* Make sure, that the file name does not start with a file separator */
+      fileName = maybeFilename.get().replaceFirst("^[\\\\/]", "");
```
johanneshiry, line 54: Pls consider addressing this sonarqube code smell by defining a private constant for the replacement regex.

johanneshiry (changes requested):

johanneshiry (changes requested):
In general this is a very good and well documented PR. 👍  
In addition to my questions/comments, pls take care of the new security hotspots shown by sonarqube and add a expressive comment when addressing them.

docs/javadoc/allclasses-frame.html
```diff
@@ -3,7 +3,7 @@
 <html lang="de">
 <head>
 <!-- Generated by javadoc -->
-<title>All Classes (PowerSystemDataModel 1.1-SNAPSHOT API)</title>
+<title>All Classes (PowerSystemDataModel 2.0.0 API)</title>
```
johanneshiry, line 6: This is invalid, as this is actually a SNAPSHOT. We might have to modify the finalize task here. Pls consult me for further discussion.

johanneshiry: The longer I think about this, the more I would prefer just keeping it that way and don't publish javadocs for snapshot builds. Hence we can keep the versioning for now but we might consider removing the java doc generation from the gradle `finalizePR` task (and adapt the CONTRIBUTING.MD accordingly). Furthermore we could discuss about having one special branch for javadocs only and automatically deploy new javadocs on each release (compare https://vaadin.com/blog/host-your-javadoc-s-online-in-github) via jenkins. This of course would not produce snapshot javadocs anymore but I think we don't need them at all. In addition this would not allow us to review javadocs anymore but tbh I would assume none of us ever has reviewed javadocs on detail (especiall not if they are 4k+ of lines ...)

ckittl:
I addressed all your comments and also improved on the security issues.

johanneshiry (approved):
👍   thanks for addressing reviewing comments. LGTM now. Thanks for the well documentation!

PR #221, opened 2020-10-14
IdCoordinateSource: Added methods to get all coordinates and the nearest coordinates

### LINKED ISSUE #218: IdCoordinateSource: Add methods to get all/ the nearest n points

mia-krause:
resolves #218

- Introduced a CoordinateDistance to wrap two coordinates and their distance, especially for return types. This class implements Comparable by comparing the distances.
- Added two methods to IdCoordinateSource: One to get all available coordinates and one to get the n nearest coordinates of a given points. These methods are useful for the IDW algorithm for weather data.
- Added a method to GridAndGeoUtils that returns a sorted set of CoordinateDistances between a base point and a collection of other coordinates, which is useful for the mentioned nearest coordinate search.

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
@@ -7,12 +7,15 @@ package edu.ie3.datamodel.io.source;
 
 import edu.ie3.datamodel.utils.CoordinateDistance;
 import edu.ie3.datamodel.utils.GridAndGeoUtils;
-import java.util.Collection;
-import java.util.List;
-import java.util.SortedSet;
+import java.util.*;
 import java.util.stream.Collectors;
 import org.locationtech.jts.geom.Point;
 
+/**
+ * This class serves mapping purposes between the ID of a coordinate and the actual coordinate with
+ * latitude and longitude values, which is especially needed for data source that don't offer
+ * combined primary or foreign keys.
+ */
 public interface IdCoordinateSource extends DataSource {
 
   /**
@@ -21,7 +24,7 @@ public interface IdCoordinateSource extends DataSource {
    * @param id the ID to look up
    * @return matching coordinate
    */
-  Point getCoordinate(Integer id);
+  Optional<Point> getCoordinate(int id);
 
   /**
    * Get the matching coordinates for the given IDs
@@ -29,7 +32,7 @@ public interface IdCoordinateSource extends DataSource {
    * @param ids the IDs to look up
    * @return the matching coordinates
    */
-  Collection<Point> getCoordinates(Integer... ids);
+  Collection<Point> getCoordinates(int... ids);
 
   /**
    * Get the ID for the coordinate point
@@ -37,7 +40,7 @@ public interface IdCoordinateSource extends DataSource {
    * @param coordinate the coordinate to look up
    * @return the matching ID
    */
-  Integer getId(Point coordinate);
+  Optional<Integer> getId(Point coordinate);
 
   /**
    * Returns all the coordinates of this source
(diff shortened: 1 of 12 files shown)


mia-krause:
!test

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
 */
 package edu.ie3.datamodel.io.source;
 
+import edu.ie3.datamodel.utils.CoordinateDistance;
+import edu.ie3.datamodel.utils.GridAndGeoUtils;
 import java.util.Collection;
+import java.util.List;
+import java.util.SortedSet;
+import java.util.stream.Collectors;
 import org.locationtech.jts.geom.Point;
```
johanneshiry, line 15: Please add javadocs about the purpose of this source.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
...
+  /**
+   * Returns the nearest n coordinate points to the given coordinate
+   *
+   * @param coordinate the coordinate to look up the nearest neighbours for
+   * @param n how many neighbours to look up
+   * @return the n nearest coordinates to the given point
+   */
+  default List<CoordinateDistance> getNearestCoordinates(Point coordinate, Integer n) {
+    SortedSet<CoordinateDistance> sortedDistances =
+        GridAndGeoUtils.getCoordinateDistances(coordinate, getAllCoordinates());
```
johanneshiry, line 58: I'm a little bit afraid that this could become a performance bottleneck if you need to query `getAllCoordinates()` every time. Is this really necessary? Maybe you could consider providing something like the example below. This would allow passing in already queried coordinates. Does this makes sense in your scenario? If you take this solution pls add and adapt javadocs accordingly. 

```java
    default List<CoordinateDistance> getNearestCoordinates(Point coordinate, int n) {
        return getNearestCoordinates(coordinate, n, Collections.emptySet());
    }

    default List<CoordinateDistance> getNearestCoordinates(Point coordinate, int n, Collection<Point> allCoordinates) {
        SortedSet<CoordinateDistance> sortedDistances = GridAndGeoUtils.getCoordinateDistances(coordinate,
                        (allCoordinates == null || allCoordinates.isEmpty()) ? getAllCoordinates() : allCoordinates);
        return sortedDistances.stream().limit(n).collect(Collectors.toList());
    }
}
```

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
...
+   */
   Collection<Point> getCoordinates(Integer... ids);
 
+  /**
+   * Get the ID for the coordinate point
+   *
+   * @param coordinate the coordinate to look up
+   * @return the matching ID
+   */
   Integer getId(Point coordinate);
```
johanneshiry, line 40: Why do you use `Integer` instead of `int`? You might consider using primitives instead of boxed parameters.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
...
+  Collection<Point> getAllCoordinates();
+
+  /**
+   * Returns the nearest n coordinate points to the given coordinate
+   *
+   * @param coordinate the coordinate to look up the nearest neighbours for
+   * @param n how many neighbours to look up
+   * @return the n nearest coordinates to the given point
+   */
+  default List<CoordinateDistance> getNearestCoordinates(Point coordinate, Integer n) {
```
johanneshiry, line 56: Why do you use `Integer` instead of `int`? You might consider using primitives instead of boxed parameters.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
...
 
 public interface IdCoordinateSource extends DataSource {
 
+  /**
+   * Get the matching coordinate for the given ID
+   *
+   * @param id the ID to look up
+   * @return matching coordinate
+   */
   Point getCoordinate(Integer id);
```
johanneshiry, line 24: Why do you use `Integer` instead of `int`? You might consider using primitives instead of boxed parameters.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -5,14 +5,57 @@
...
 
+  /**
+   * Get the matching coordinates for the given IDs
+   *
+   * @param ids the IDs to look up
+   * @return the matching coordinates
+   */
   Collection<Point> getCoordinates(Integer... ids);
```
johanneshiry, line 32: Why do you use `Integer` instead of `int`? You might consider using primitives instead of boxed parameters.

src/main/java/edu/ie3/datamodel/utils/GridAndGeoUtils.java
```diff
@@ -156,4 +155,19 @@ private static Point buildSafePoint(Point p1) {
...
+   *
+   * @param baseCoordinate the base point
+   * @param coordinates the points to calculate the distance from the base point for
+   * @return a sorted set of distances between the base and other coordinates
+   */
+  public static SortedSet<CoordinateDistance> getCoordinateDistances(
+      Point baseCoordinate, Collection<Point> coordinates) {
+    return coordinates.stream()
+        .map(coordinate -> new CoordinateDistance(baseCoordinate, coordinate))
+        .collect(Collectors.toCollection(TreeSet::new));
```
johanneshiry, line 171: 👍  for using TreeSet here

johanneshiry (changes requested):
Thanks for your work. I added some questions and remarks I would prefer being resolved before merging.

Furthermore pls remember addressing the following: 
[ ] solve all codacy issues
[ ] adapt CHANGELOG.MD and add your work

mia-krause:
Is there a way I can mark codacy issues as "resolved" or similar? Otherwise I can't solve it this time, as it is the equals / == issue I mentioned

johanneshiry:
> Is there a way I can mark codacy issues as "resolved" or similar? Otherwise I can't solve it this time, as it is the equals / == issue I mentioned

solved it - just ignore this one for now - thanks

johanneshiry:
!test

johanneshiry (approved):

PR #224, opened 2020-10-15
Added SQL and Couchbase weather source

### LINKED ISSUE #223: Add Weather Sources

mia-krause:
~Resolves~ #223  (and has already merged #221)


src/main/java/edu/ie3/datamodel/io/connectors/CouchbaseConnector.java
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Implements a DataConnector for Couchbase. Couchbase is a JSON document based database. <br>
+ * Entities will be persisted in documents which are then mapped by a key.
+ */
+public class CouchbaseConnector implements DataConnector {
+
+  private static final String DEFAULT_URL = "127.0.0.1";
+  private static final String DEFAULT_BUCKET_NAME = "ie3_in";
+  private static final String DEFAULT_USERNAME = "Administrator";
+  private static final String DEFAULT_PASSWORD = "password";
```
johanneshiry, line 27: Does it make sense to provide default credentials here? Don't get me wrong but IMHO it doesn't really make sense because they are barely useful in a real world scenario aren't they? In addition, having a default constructor with these parameters might lead to more confusion than it helps. Therefore I would always prefer to enforce the user to provide credentials to made them obvious instead of hiding them in a default constructor ...

src/main/java/edu/ie3/datamodel/io/connectors/CouchbaseConnector.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  /**
+   * Initializes a new CouchbaseConnector, uses the default URL ({@value #DEFAULT_URL}), bucket name
+   * ({@value #DEFAULT_BUCKET_NAME}), user name ({@value #DEFAULT_USERNAME}) and password ({@value
+   * #DEFAULT_PASSWORD})
+   */
+  public CouchbaseConnector() {
+    this(DEFAULT_URL, DEFAULT_BUCKET_NAME, DEFAULT_USERNAME, DEFAULT_PASSWORD);
+  }
```
johanneshiry, line 62:
```suggestion
```

src/main/java/edu/ie3/datamodel/io/connectors/CouchbaseConnector.java
johanneshiry: btw this holds true for alle existing connectors we have so far. I think we already have some with default creds available that already confused me a few days ago :D

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * Initializes a SqlConnector with default parameters: <br>
+   * jdbcUrl = jdbc:postgresql://localhost:5432/simona <br>
+   * username = postgres <br>
+   * password = root <br>
+   */
+  public SqlConnector() {
+    this("jdbc:postgresql://localhost:5432/simona", "postgres", "root");
+  }
```
johanneshiry, line 47: same as in `CouchbaseConnector` - would remove this and make the actual parameters explicit when calling the parametrized constructor

```suggestion
```

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Establishes and returns a database connection
+   *
+   * @return the connection object
+   * @throws SQLException if the connection could not be establlished
+   */
+  public Connection getConnection() throws SQLException {
+    Properties connectionProps = new Properties();
+    connectionProps.put("user", userName);
+    connectionProps.put("password", password);
```
johanneshiry, line 90: Properties might become a final field and the put operations could be done in the constructor as they are not changed afterwards as far as I can observe

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -0,0 +1,143 @@
...
+    try {
+      return DriverManager.getConnection(jdbcUrl, connectionProps);
+    } catch (SQLException e) {
+      throw new SQLException("Could not establish connection: ", e);
+    }
+  }
+
+  @Override
+  public void shutdown() {
+    // Nothing needs to be closed or shutdown
```
johanneshiry, line 100: Is this bc we're not managing connections? Otherwise I would assume closing the connections would make sense? 🤔

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.stream.Stream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.locationtech.jts.geom.Point;
+
+/** Couchbase Source for weather data */
+public class CouchbaseWeatherSource implements WeatherSource {
+  private static final Logger logger = LogManager.getLogger(CouchbaseWeatherSource.class);
+  private static final String COORDINATE_ID_COLUMN_NAME = "coordinate";
+  private static final String TIMESTAMP_PATTERN = "yyyy-MM-dd'T'HH:mm:ssxxx";
```
johanneshiry, line 37: Same - pls don't hard code important parameters but le the user provide them in a parametrized constructor. 
As the timestamp pattern is indeed a "classic" one, here a second constructor with this default pattern IMHO makes sense tough.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Generates a key for weather documents with the pattern: {@code
+   * weather::<coordinate_id>::<time>}
+   *
+   * @param time the timestamp for the weather data
+   * @param coordinate the coordinate of the weather data
+   * @return a weather document key
+   */
+  public String generateWeatherKey(ZonedDateTime time, Point coordinate) {
+    String key = "weather::";
```
johanneshiry, line 114: is this always the same? Otherwise pls prevent hardcoding it, but make it a final field and add it to a parametrized constructor

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -0,0 +1,271 @@
...
+/** SQL source for weather data */
+public class SqlWeatherSource implements WeatherSource {
+  private static final Logger mainLogger = LogManager.getLogger("Main");
+
+  private static final String DEFAULT_WEATHER_TABLE = "weather";
+  private static final String DEFAULT_WEATHER_SCHEMA = "public";
+  private static final String BASIC_QUERY =
+      "SELECT * FROM " + DEFAULT_WEATHER_SCHEMA + "." + DEFAULT_WEATHER_TABLE;
+  private static final String DEFAULT_COORDINATE_COLUMN = "coordinate";
+  private static final String DEFAULT_TIME_COLUMN = "time";
```
johanneshiry, line 38: Same as in other sources. Pls avoid hard coded parameters.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -0,0 +1,271 @@
...
+  /**
+   * Initializes a new SqlWeatherSource
+   *
+   * @param connector the connector needed for database connection
+   * @param idCoordinateSource a coordinate source to map ids to points
+   */
+  public SqlWeatherSource(SqlConnector connector, IdCoordinateSource idCoordinateSource) {
+    this.connector = connector;
+    this.idCoordinateSource = idCoordinateSource;
+    this.weatherFactory = new TimeBasedWeatherValueFactory("yyyy-MM-dd HH:mm:ss.S");
```
johanneshiry, line 55: is this pattern always the same? Otherwise avoid hardcoding it pls

johanneshiry (changes requested):
Thank you very much for your work.
I added some important remarks and change request. If you want to pls consult me for a short call before addressing it. 

Furthermore I would like to know why you chose classic "fetch" implementation instead of using a streams implementation (e.g. [akka with slick](https://doc.akka.io/docs/alpakka/current/slick.html) or [alpakka couchbase](https://doc.akka.io/docs/alpakka/current/couchbase.html)?)

Please also remember, that the following needs to be done:

- [ ] adapt CHANGELOG.MD and add your work

ckittl:
I edited your description, as it doesn't resolve the issue entirely (still a csv weather source would be missing).

Please consider either splitting up the issue or addressing this type of source as well. If you need any help, please just let me know. There are already foundations available as stated in #223.

src/main/java/edu/ie3/datamodel/io/connectors/CouchbaseConnector.java
```diff
@@ -0,0 +1,139 @@
...
+    return keys.stream().map(key -> session.async().get(key)).collect(Collectors.toList());
+  }
+
+  /**
+   * Returns a future for search result of the given key
+   *
+   * @param key the key to get the document for
+   * @return future of the search result
+   */
+  public CompletableFuture<GetResult> get(String key) {
```
johanneshiry, line 126: :+1: for using `CompletableFuture`

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
johanneshiry: This still holds true. Pls either comment why `Properties` cannot become a final field or make it a final field.

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
johanneshiry: same here: what's up with this comment?

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.locationtech.jts.geom.Point;
+
+/** Couchbase Source for weather data */
+public class CouchbaseWeatherSource implements WeatherSource {
+  private static final Logger logger = LogManager.getLogger(CouchbaseWeatherSource.class);
+  private static final String DEFAULT_TIMESTAMP_PATTERN = "yyyy-MM-dd'T'HH:mm:ssxxx";
```
johanneshiry, line 36: Is there a reason why you don't use the default pattern used in `edu.ie3.util.TimeUtil`? This might make sense for consistency ...?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -0,0 +1,341 @@
...
+import java.util.stream.Stream;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.locationtech.jts.geom.Point;
+
+/** SQL source for weather data */
+public class SqlWeatherSource implements WeatherSource {
+  private static final Logger logger = LogManager.getLogger(SqlWeatherSource.class);
+
+  private static final String DEFAULT_TIMESTAMP_PATTERN = "yyyy-MM-dd HH:mm:ss.0";
```
johanneshiry, line 34: Is there a reason why you don't use the default pattern used in edu.ie3.util.TimeUtil? This might make sense for consistency ...?

johanneshiry (changes requested):
Thank you very much for your work. In general looks good to me. Before merging I would prefer having the conversations and questions resolved as well I would kindly ask you again to add your work to the CHANGELOG.md. 

Thank you :-)

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
mia-krause: Because this is the default pattern of the database, so any return strings will have this pattern, unless explicitly changed.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
mia-krause: Same here

mia-krause:
I weirdly can't reply to some conversations so here are the other points...

- The properties cannot be final, as the only possible parameter to a constructor is another Property object and no singular parameters.
- slick / alpakka: I wanted the sources to be generic and not just for use with akka
- Comments on shutdown methods: Nothing needs to be shutdown, as we use short-lived sessions, that are autoclosable and wrapped in a try-with-resources for every query.

johanneshiry (approved):

PR #231, opened 2020-10-27
Properly considering directory hierarchy when instantiating csv writers

### LINKED ISSUE #230: Fix csv file connector together with folder hierarchy
Currently, there is a slight problem using the csv file connector to init a buffered writer in conjunction with a specified directory hierarchy. As parts of the hierarchy are handed in relatively to the base path, the "file name" does contain parts of information about the final directory hierarchy.

This is currently not considered thoroughly when creating the missing directories in the files absolute path.

ckittl:
Fixes #230


src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -79,12 +106,12 @@ protected final void writeFileHeader(String[] headLineElements) throws IOExcepti
    * @param entries Entries to write to the line of the file
    * @throws IOException If writing is not possible
    */
-  protected final void writeOneLine(String[] entries) throws IOException {
+  private final void writeOneLine(String[] entries) throws IOException {
```
johanneshiry, line 109: either make this method `protected` again or remove the `final` keyword as private methods are always implicitly `final`

src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -110,18 +137,26 @@ public Writer append(char c) {
   @Override
   public boolean equals(Object o) {
```
johanneshiry, line 138: Why this change in `equals()`?

ckittl: Bc the outline of the classes attributes has changed (unpacking the information from former `CsvFileDefinition`, that has been the only attribute).

johanneshiry (approved):
LGTM 👍

PR #233, opened 2020-10-30
Harmonized result file naming and time columns

### LINKED ISSUE #149: Harmonized file naming for result entities
The default file naming strategy derives the file names from model classes' names.

Currently, for input models, the class name is transferred from camel to snake case, whereas this does not apply to the result model naming.

Please adapt `FileNamingStrategy#buildResultEntityString(Class<? extends ResultEntity>)` accordingly. It should also return an Optional as it's sibblings.

### LINKED ISSUE #175: Rename all time based columns to "time"
This should be done for three reasons:

1. They represent a time, so they should be named as such
2. It's more intuitive if all columns on all models representing the same field, have the same value
3. InfluxDB requires the time based field to be named "time"

However, as this is a huge change in the existing data, this change should be paused until the next release.

ckittl:
Closes #175 and fixes #149


johanneshiry:
!test

johanneshiry (approved):
LGTM 👍

johanneshiry:
!test

Issue #182, opened 2020-07-31
Concurrency issue in CsvSink


ckittl:
It seems, that there is a sort of concurrency issue to the `CsvSink`. I observe the following mentioned issues, when passing models to `CsvSink#persistAllIgnoreNested(Collection<C>)`.

### Observation 1
When passing in several objects of the same type, an warning is raised, that the content will be appended to an already existing file with the targeted name. However, such a file has not been apparent before.

**Possible explanation**
The whole collection of elements is treated concurrently (cf. [here](https://github.com/ie3-institute/PowerSystemDataModel/blob/bbc85c273efc541e71e435849d5d07a33383574f/src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java#L142)). For each element, it is checked, if there is a writer available for the respective class, yet. If not, it is initialized. However, it could occur, that the first element doesn't find a matching writer and starts to initialize one, while at the same time element two of the same class also doesn't see the writer (initialization started, but not finished yet, but a file has already been created), therefore also initializing a second writer. As the writers are registered in a map, the first one is abandoned after writer the first element.

**Proposed solution**
On issuing `CsvSink#persistAllIgnoreNested(Collection<C>)`, split up the given collection into sub collections only holding elements of one class. Those sub collections can be treated concurrently (not the elements within it). The writers are anyways the place, where concurrency ends. Therefore, concurrent attempts to write to the same file would be avoided at a very early stage. @johanneshiry: What's your opinion on that? I think the sink was your concept.

### Observation 2
Now and then not all passed elements actually appear in the csv files. Actually I don't have any clue yet, why this is happening. Maybe it's a flushing issue?

**Proposed next steps / Questions to answer**

- Trying to evaluate the problem a bit better.
- Might the calling thread be exited, before the flushing could finish?
- Think, if the concept under observation could help to mitigate the problem?
  - We `flush()` after each line, what subverts the concept of a buffered writer. With above concept, it could be called, after all elements of the same class have been written.

PR #238, opened 2020-11-04
Make getting a writer and writing an entry to csv thread safe

### LINKED ISSUE #235: Resolve race condition on writers
The `CsvFileConnector` holds a map from target class to corresponding buffered writer. However, when accessing this map in a concurrent context like `CsvFileSink#persistAllIgnoreNested(Collection)`, the different threads try to access the entry, do not find a entry and create a new buffered writer.

Only one of those writers will win the race to the map, therefore the rest is abandoned and pollutes memory.

*Possible solution:* Synchronize the map.

https://github.com/ie3-institute/PowerSystemDataModel/blob/5ab61668bd29c155753ed57b3190592291378961/src/main/java/edu/ie3/datamodel/io/connectors/CsvFileConnector.java#L38-L39

ckittl:
Resolves #235

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/connectors/CsvFileConnector.java
+++ b/src/main/java/edu/ie3/datamodel/io/connectors/CsvFileConnector.java
@@ -48,7 +48,7 @@ public class CsvFileConnector implements DataConnector {
     this.fileNamingStrategy = fileNamingStrategy;
   }
 
-  public BufferedCsvWriter getOrInitWriter(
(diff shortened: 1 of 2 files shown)


johanneshiry:
Does this relate to #182

ckittl:
> Does this relate to #182?

At least for observation 1, not sure if it solves the second one as well.

johanneshiry (approved):

Issue #248, opened 2020-11-11
Fix invalid InfluxDbSink session handling


johanneshiry:
The current implementation of the `InfluxDbSinks` has several architectural issues that causes a huge performance drop when executing queries on the database. Especially in a concurrent environment with several thousand write calls the existing implementation is not usable at all. Furthermore, the current implementation prevents tuning performance relevant parameters.

Hence some major refactoring is required incl. changing method signatures. 

- [x] adapt `InfluxDbConnector` to hold only one session per connector and not creating a new session for each call on `getSession()`
- [x] provide constructors for `InfluxDbConnector` that allows for passing in InfluxDb session tuning parameters (especially `BatchOptions`)
- [x] remove `null` for ` scenarioName in `InfluxDbConnector` constructor (relates to #247)
- [x] shutdown session in `InfluxDbConnector#shutdown()`

PR #249, opened 2020-11-12
Resolve influxDBConnector issues 

### LINKED ISSUE #247: InfluxDbSink throws NPE
The current implementation of the `InfluxDbSink` tries to retrieve a scenario name in `InfluxDbSink:139` from the provided `InfluxDbConnector`. However, one of the default `InfluxDbConnector` constructors passes in a `null` as scenario parameter which causes a NullPointerException. Furthermore, the documentation of this constructor says that one should not use it at all for persisting (????). Therefore I would prefer removing it but maybe @mia-krause can help with an explanation about the purpose of this constructor...?! 

https://github.com/ie3-institute/PowerSystemDataModel/blob/1bea36c08c7180a9454d2da1a2f1822a364f9d6b/src/main/java/edu/ie3/datamodel/io/connectors/InfluxDbConnector.java#L62

https://github.com/ie3-institute/PowerSystemDataModel/blob/1bea36c08c7180a9454d2da1a2f1822a364f9d6b/src/main/java/edu/ie3/datamodel/io/sink/InfluxDbSink.java#L139

johanneshiry:
Resolves #247  and #248 

@mia-krause in contrast to the existing implementation I keep the session open once it has been available to a `InfluxDbConnector` instead of creating a new on for each `#getSession()` call. . Based on the influxdb-java docs this is the recommended way to do. Reason for this is that the old implementation of `#getSession()` created a massive overhead when running concurrent writing operations as well as the application of the batch size was useless, bc for `#persist(C entity)` each session has been used exactly for one call and then closed. 

I don't know if this breaks with one of your concepts but might you could provide some insights why you initially wanted to use a new session for each persist call?


src/main/java/edu/ie3/datamodel/io/connectors/InfluxDbConnector.java
```diff
@@ -31,43 +31,72 @@
...
+   * Initializes a new InfluxDbConnector with the given url, databaseName and scenario name.
    *
    * @param url the connection url for the influxDB database
    * @param databaseName the name of the database to which the connection should be established
+   * @param scenarioName the name of the simulation scenario which will be used in influxDB
+   *     measurement names
    */
-  public InfluxDbConnector(String url, String databaseName) {
-    this(url, databaseName, null);
+  public InfluxDbConnector(String url, String databaseName, String scenarioName) {
```
mia-krause, line 88: The scenario name was meant to be used for output data, to reference the used scenario. If we use only weather, we don't need the scenario name. If we require it, would it make sense to offer a default constant? 
Either an empty scenario `public static final String NO_SCENARIO = "no_scenario";` or a weather specific one `public static final String WEATHER_SCENARIO = "weather_scenario";`.

johanneshiry:
!test

johanneshiry:
!test

johanneshiry:
!test

mia-krause (dismissed):

johanneshiry:
!test

mia-krause (approved):

PR #257, opened 2020-12-04
DWD ICON model connector

### LINKED ISSUE #256: DWD Icon-connector
Implement csv and SQL connector to connect to home-baked DWD Icon data model.

ckittl:
Resolves #256


src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+    java.util.UUID uuid = java.util.UUID.randomUUID();
+    ZonedDateTime time = timeUtil.toZonedDateTime(data.getField(TIME));
+    ComparableQuantity<Irradiation> directIrradiation =
+        data.getQuantity(DIRECT_IRRADIATION, PowerSystemUnits.WATT_PER_SQUAREMETRE)
+            .to(StandardUnits.IRRADIATION);
+    ComparableQuantity<Irradiation> diffuseIrradiation =
+        data.getQuantity(DIFFUSE_IRRADIATION, PowerSystemUnits.WATT_PER_SQUAREMETRE)
+            .to(StandardUnits.IRRADIATION);
+    ComparableQuantity<Temperature> temperature =
+        data.getQuantity(TEMPERATURE, PowerSystemUnits.KELVIN).to(StandardUnits.TEMPERATURE);
```
johanneshiry, line 117:
```suggestion
        data.getQuantity(TEMPERATURE, Units.KELVIN).to(StandardUnits.TEMPERATURE);
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+            DIRECT_IRRADIATION,
+            TEMPERATURE,
+            WIND_VELOCITY_U,
+            WIND_VELOCITY_V);
+    Set<String> allParameters = new HashSet<>(minConstructorParams);
+    allParameters.addAll(
+        newSet(
+            "albRad",
+            "asobS",
+            "aswdifdS",
```
johanneshiry, line 76:
```suggestion
            DIFFUSE_IRRADIATION,
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+            "aswdifuS",
+            "aswdirS",
```
johanneshiry, line 78:
```suggestion
            DIRECT_IRRADIATION,
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+            "t2m",
+            "tG",
+            "u10m",
+            "u131m",
```
johanneshiry, line 82:
```suggestion
            WIND_VELOCITY_U,
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+            "u20m",
+            "u216m",
+            "u65m",
+            "v10m",
+            "v131m",
```
johanneshiry, line 87:
```suggestion
            WIND_VELOCITY_V,
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+   * @return A {@link ComparableQuantity} of type {@link Speed}, that is converted to {@link
+   *     StandardUnits#WIND_VELOCITY}
+   */
+  private static ComparableQuantity<Speed> getWindVelocity(TimeBasedWeatherValueData data) {
+    /* Get the three dimensional parts of the wind velocity vector in cartesian coordinates */
+    double u = data.getDouble(WIND_VELOCITY_U);
+    double v = data.getDouble(WIND_VELOCITY_V);
+
+    double velocity = Math.sqrt(Math.pow(u, 2) + Math.pow(v, 2));
+    return Quantities.getQuantity(velocity, PowerSystemUnits.METRE_PER_SECOND)
```
johanneshiry, line 174:
```suggestion
    return Quantities.getQuantity(velocity, Units.METRE_PER_SECOND)
```

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+   */
+  private static ComparableQuantity<Angle> getWindDirection(TimeBasedWeatherValueData data) {
+    /* Get the three dimensional parts of the wind velocity vector in cartesian coordinates */
+    double u =
+        data.getDouble(WIND_VELOCITY_U); // Wind component from west to east (parallel to latitudes)
+    double v =
+        data.getDouble(
+            WIND_VELOCITY_V); // Wind component from south to north (parallel to longitudes)
+
+    double angle = Math.toDegrees(Math.atan2(-u, -v));
```
johanneshiry, line 152: Just curiosity bc I'm not familiar with the assumptions this is based on: why do we need the two minus signs here?

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
```diff
@@ -0,0 +1,177 @@
...
+            directIrradiation,
+            diffuseIrradiation,
+            temperature,
+            windDirection,
+            windVelocity);
+    return new TimeBasedValue<>(uuid, time, weatherValue);
+  }
+
+  /**
+   * Determines the wind direction. In ICON the wind velocity is given in three dimensional
```
johanneshiry, line 132: 🥇  for the and comprehensive explanation

src/main/java/edu/ie3/datamodel/models/StandardUnits.java
```diff
@@ -113,7 +113,7 @@
   /** Standard unit for reactive power characteristics */
   public static final Unit<Dimensionless> Q_CHARACTERISTIC = PU;
   /** Solar irradiation on a flat surface */
-  public static final Unit<Irradiation> IRRADIATION = KILOWATTHOUR_PER_SQUAREMETRE;
+  public static final Unit<Irradiation> IRRADIATION = KILOWATT_PER_SQUAREMETRE;
```
johanneshiry, line 116: This is breaking and should be marked correspondingly in CHANGELOG.md

johanneshiry (changes requested):
I really like the split up of the value factories and how they are passed into the sources. This preserves the flexibility we have in a very good way. 💯 

I added some minor change requests for stylistic reasons you might consider addressing.

Furthermore I would ask you to please 
- fix the javadoc to let CI check your work (will review afterwards again) 
- adapt CHANGELOG.md accordingly as you added some breaking changes at several places which should be noted (not only in the StandardUnits, but also by changing constructor signatures etc.) - this is btw still okay as we intend to release a breaking version (2.0) soon and hence if marked accordingly should be acceptable.

ckittl:
I'm not ready, yet. 😅

johanneshiry:
> I'm not ready, yet. 😅

😅  - oh ... sorry my fault - so much enthusiasm on my side that I overlooked the draft 🙈

src/main/java/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactory.java
ckittl: This is also due to the fact, that the wind velocity vector points in the direct opposite direction as the vector of the wind velocity.

ckittl:
This PR is ready, yet. But I feel like the sources need some rework in order to maintain consistent logic. Maybe, we can use the place here to have some discussions about this, @johanneshiry @mia-krause:

Currently, in csv files we expect the column names to be given in snake case and we convert them into camel case in runtime to build up the java models. However, in all other sources, this is not explicitly addressed, therefore seemingly expecting camel case on the first hand in actual data sources.

I feel like this is very error prone, as I experienced during this work. In order to fit the column scheme of DWD Icon model, we hold at ie³, I also implemented the assumption, that the SQL weather tables have their columns specified as snake case. From my point of view we should do one of both things:

1. Set up a general requirement to have all column or key specifications in sources in either camel oder snake case or
2. Implement all sources that adaptive, that they can be configured or detect the case themselves.

What are your thoughts about it? I have a slight tendency towards option 1, whereas I'm not totally sure, if this is possible with all yet implemented types of sources and if this complies with their conventions. I think here you, @mia-krause, are the expert.

As of the pending design decisions, there are no integration tests implemented for ICON column scheme, despite csv and SQL, which I feel are the most important right now. Depending of the result of our discussion, I'm willing to provide corresponding issue, so that we can work on this.

johanneshiry:
!test

johanneshiry (dismissed):
well crafted and tested work 🥇

johanneshiry (approved):

johanneshiry (approved):

PR #259, opened 2020-12-10
jh/#258 snake case headline db source restriction

### LINKED ISSUE #258: Restrict weather input data headlines to snake case
As discovered by @ckittl there is no mechanism inside the current database sources for weather data that enforces column headlines to be in snake case. While during the initialization this is not an issue, it might cause issues when queries are executed to receive specific weather data, because the queries assume the headlines to be in snake case. 

To enforce snake case in the database sources a headline check should be executed in the constructor of the database sources using the fields in the corresponding factory.

johanneshiry:
resolves #258 

In contrast to the initially intended restriction, this PR ensures for SQLWeatherSource that the queries are build with respect to the actual column names in the database as a restriction to snake case only is not that easy to pursue. 
As we currently don't have ICON weather data support for couchbase and influxdb, both corresponding sources are ignored. However, influxdb has been adapted to actually be able to handle ICON weather data as well.


src/main/java/edu/ie3/datamodel/io/factory/Factory.java
```diff
@@ -173,4 +173,19 @@ protected static StringBuilder getFieldsString(Set<String>... fieldSets) {
     set.addAll(Arrays.asList(attributes));
     return set;
   }
+
+  /**
+   * Expands a set of attributes with further attributes. This method should always be used when
+   * returning attribute sets, i.e. through getting the needed fields.
```
ckittl, line 179:
```suggestion
   * Expands a set of attributes with further attributes. This method should always be used when
   * returning attribute sets, i.e. through getting the needed fields. The set maintains a lexicographic
   * order, that is case-insensitive.
```

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -146,6 +159,40 @@ public SqlWeatherSource(
     return Optional.of(timeBasedValues.get(0));
   }
 
+  /**
+   * Determine the corresponding the database column name based on the provided factory field
```
ckittl, line 163:
```suggestion
   * Determine the corresponding database column name based on the provided factory field
```

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -146,6 +159,40 @@ public SqlWeatherSource(
...
+   * parameter name. Needed to support camel as well as snake case database column names.
+   *
+   * @param factoryColumnName the name of the field parameter set in the entity factory
+   * @param connector the sql connector of this source
+   * @param weatherTableName the table name where the weather is stored
+   * @return the column name that corresponds to the provided field parameter or an empty optional
+   *     if no matching column can be found
+   */
+  private Optional<String> dbColumnName(
```
ckittl, line 172: I would prefer a method name with a verb in it. E.g.
```suggestion
  private Optional<String> determineDbColumnName(
```

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -146,6 +159,40 @@ public SqlWeatherSource(
...
+        String databaseColumnName = rs.getString("COLUMN_NAME");
+        if (StringUtils.snakeCaseToCamelCase(databaseColumnName)
+            .equalsIgnoreCase(factoryColumnName)) {
+          dbColumnName = Optional.of(databaseColumnName);
+          break;
+        }
+      }
+
+    } catch (SQLException throwables) {
+      throwables.printStackTrace();
```
ckittl, line 191: Is there a reason to not log this exception properly (cf. vulnerability detected by SonarQube)?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -62,26 +63,38 @@ public SqlWeatherSource(
...
+
+    String dbCoordColumnName =
+        dbColumnName(weatherFactory.getCoordinateIdFieldString(), connector, weatherTableName)
+            .orElseThrow(
+                () ->
+                    new InvalidWeatherColumnNameException(
+                        "Cannot find column for '"
+                            + weatherFactory.getCoordinateIdFieldString()
+                            + "' in provided weather data configuration."
+                            + "Please ensure that the database connection is working and the column names are correct!"));
```
ckittl, line 87:
```suggestion
    String dbCoordColumnName =
        dbColumnName(factoryCoordinateFieldName, connector, weatherTableName)
            .orElseThrow(
                () ->
                    new InvalidWeatherColumnNameException(
                        "Cannot find column for '"
                            + factoryCoordinateFieldName
                            + "' in provided weather data configuration."
                            + "Please ensure that the database connection is working and the column names are correct!"));
```

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -62,26 +63,38 @@ public SqlWeatherSource(
       TimeBasedWeatherValueFactory weatherFactory) {
     this.connector = connector;
     this.idCoordinateSource = idCoordinateSource;
-    this.coordinateColumnName = weatherFactory.getCoordinateIdFieldString();
     this.weatherFactory = weatherFactory;
+    this.factoryCoordinateFieldName = weatherFactory.getCoordinateIdFieldString();
 
-    String timeColumnName = weatherFactory.getTimeFieldString();
+    String dbTimeColumnName =
+        dbColumnName(weatherFactory.getTimeFieldString(), connector, weatherTableName)
```
ckittl, line 70: Possibly introduce attribute for `weatherFactory.getTimeFieldString()` as well?

ckittl (changes requested):
Despite of the already mentioned vulnerability in not logging the stack trace of the sql exception, the coverage is to low to pass quality gate. Maybe you have a chance to make this pass?

ckittl:
!test

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
johanneshiry: nope - looks like I missed adapting the auto-generated code for try-catch

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
johanneshiry: good catch - thx 👍🏼

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
johanneshiry: IMHO not required here, as the second call only takes place on error. Hence in normal operation there is only one call required which does not need an attribute here.

ckittl (approved):

Issue #271, opened 2021-02-03
Minor database issues


mia-krause:
- `InfluxDbWeatherSource` builds a bad query string, by querying for time first, which is not indexed, instead of tag first
- `SqlWeatherSource` does not repurpose connection, leading to issues with too many open sessions

johanneshiry:
@mia-krause / @ckittl is this addressed?

ckittl:
Yes.

PR #272, opened 2021-02-03
Changed order of fields in InfluxDB queries, repurpose SQl connection

mia-krause:
Solves #271


ckittl:
!test

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -172,29 +172,34 @@ public InfluxDbWeatherSource(
 
   private String createQueryStringForIntervalAndCoordinate(
```
ckittl, line 173:
```suggestion
  private String createQueryStringForCoordinateAndTimeInterval(
```

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -172,29 +172,34 @@ public InfluxDbWeatherSource(
...
-    return createQueryStringForInterval(timeInterval)
+    return BASIC_QUERY_STRING
+        + " where "
+        + createCoordinateConstraintString(coordinateId)
         + " and "
-        + createCoordinateConstraintString(coordinateId);
+        + createTimeConstraint(timeInterval);
   }
 
   private String createQueryStringForDateAndCoordinate(ZonedDateTime date, int coordinateId) {
```
ckittl, line 182:
```suggestion
  private String createQueryStringForCoordinateAndTime(ZonedDateTime date, int coordinateId) {
```

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -172,29 +172,34 @@ public InfluxDbWeatherSource(
 
   private String createQueryStringForIntervalAndCoordinate(
       ClosedInterval<ZonedDateTime> timeInterval, int coordinateId) {
-    return createQueryStringForInterval(timeInterval)
+    return BASIC_QUERY_STRING
+        + " where "
```
ckittl, line 176: You might consider introducing a constant for `" where "`. It's not too vital, but reduces the number of code smells...

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -361,4 +361,9 @@ private String createQueryStringForTimeIntervalAndCoordinates(
     }
     return coordinateToTimeSeries;
   }
+
+  private Connection getConnection() throws SQLException {
```
ckittl, line 365: From my POV we should discuss, if holding a connection should better be moved to the connector itself. Either `SqlConnector` having an attribute `reuseConnection` or the getter being like

```java
public getConnection() throws SQLException {
  return getConnection(false);
}

public getConnection(boolean reuseConnection) throws SQLException {
  if (reuseConnection && (connection == null || connection.isClosed()) || !reuseConnection) connection = connector.getConnection();
    return connection;
}
```

What are your thoughts on this?

johanneshiry: Very good point. Especially bc otherwise it would be impossible to execute concurrent db calls with the same weather source (we would have to create a new weather source for each concurrent call). However, I would prefer having a slightly more elaborated approach, where the number of parallel connections is tracked in order to not create too much concurrent connections. This might be possible with a list or map holding the actual open connections and either returns one of them that is currently not used or creates a new one. Furthermore, in this case the connector should keep "its space clean" by automatically remove unused connections whenever a new `getConnection` call is executed.

Maybe we can discuss this in a short call?

ckittl: I'm open to a call. But then we might consider splitting up the PR into on InfluxDB- and one SQL-PR?!

johanneshiry:
!test

ckittl (dismissed):

ckittl (approved):

PR #276, opened 2021-02-09
Extracting mapping source from former time series source

### LINKED ISSUE #274: Restructure time series sources
- [ ] Split up into mapping and actual data sources
- [ ] One source per time series and not everything in one pot

ckittl:
Resolves #274


ckittl:
!test

ckittl:
What did I do here?

I modularized the time series sources a bit further:

- Having a distinct mapping source, that provides information on which time series is there for which model as well as the types of time series aka. meta information
- The time series source itself only provides information about one time series at a time (including the possibility to query results for periods in time or specific instances in time)
- The CsvTimeSeriesSource` has a factory to build correct source from given meta information

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   BREAKING: Harmonized field naming for time information
 -   BREAKING: Properly applying snake case to result file names
 -   deprecated `TarballUtils`
+-   Reworking the time series source (One source per time series, distinct mapping source, factory pattern)
```
johanneshiry, line 28:
```suggestion
-   Reworking the time series source (one source per time series, distinct mapping source, factory pattern)
```

src/main/java/edu/ie3/datamodel/utils/TimeSeriesUtil.java
```diff
@@ -0,0 +1,32 @@
...
+*/
+package edu.ie3.datamodel.utils;
+
+import edu.ie3.datamodel.models.timeseries.individual.IndividualTimeSeries;
+import edu.ie3.datamodel.models.value.Value;
+import edu.ie3.util.interval.ClosedInterval;
+import java.time.ZonedDateTime;
+import java.util.stream.Collectors;
+
+public class TimeSeriesUtil {
```
johanneshiry, line 14: I would expect this is a util class? Can be static? Private constructor is missing here as well

johanneshiry (changes requested):
Thank you very much for your contribution. 👍🏼  
Coverage is also fine for now, as mainly equals or exceptions are not tested. 
Just had a few minor remarks that should be fast addressable. :-)

src/main/java/edu/ie3/datamodel/utils/TimeSeriesUtil.java
ckittl: There is no such thing like a static outer class, as it effectively is static right from the start. :smiley:

johanneshiry:
!test

johanneshiry (approved):

PR #278, opened 2021-02-18
Replaced InvalidGridException with a logger warning

### LINKED ISSUE #148: Replace the InvalidGridException in JointGrid:50 with a logged warning
At the moment there is an `InvalidGridException` thrown whenever a `JointGridContainer` only consists of a single `SubGridContainer`. Based on the exception message (`"This joint grid model only contains one single grid. Consider using SubGridContainer."`) I would propose to replace the Exception with a warning so that one can still use a `JointGridContainer` with a single `SubGrid` even if this is not the intended by design. 

The reason for this suggestion is, that there is no real disadvantage of using a `JointGridContainer` with a single grid only but there might be some benefits when working with a `GridContainer` that can become a `JointGrid` and a `SubGrid` during runtime.

Would like to read your opinions on this issue and if you would agree that replacing the exception with a warn logging would be sufficient?

ahamshubham:
Resolves #148. Changes done -
- Replaced `InvalidGridException` with a logger warning in case a `JointGridContainer` only consists of a single `SubGridContainer`.


ahamshubham:
Since, the code coverage test is failing, I suppose I need to add some test cases to check the newly added changes.

ahamshubham:
!test

ahamshubham:
!test

ckittl (approved):
Your test actually does only add limited value. Ideally, you would also test, that the log call has been made. However, in our case it is not easy to do.

It would work, if you would have the logger non-static and non-final. Then you could replace it with a mock or a spy on a delegate. However, completeness of the test doesn't justify design changes with given drawbacks.

PR #282, opened 2021-02-25
Fix InfluxDB image version to latest working version

ckittl:
Tests with InfluxDB testcontainers referenced the latest image version on dockerhub. However, somehow something has changed, why our tests started to fail.


ckittl:
Replaces #281 due to malformed branch naming. Shame on me... :hankey:

johanneshiry (approved):

Issue #283, opened 2021-02-26
Preparations for using the data model in distributed simulations


sebastian-peter:
In order for the data model to be useful in distributed simulations, a couple of changes need to be made, e.g. making some types serializable. 
Development is ongoing, changes made are not definitive.

Edit: Current changes (making some classes serializable) is now needed in Simona.

PR #289, opened 2021-03-03
Sp/#283 distributed preparations

sebastian-peter:
Solves #283 

These changes are needed in order to continue with development of distributed computing in SIMONA

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ dependencies {
 	//	 ie³ power system utils
 	compile 'com.github.ie3-institute:PowerSystemUtils:1.5+'
 
-	implementation 'tech.units:indriya:2.0.4'
+	implementation 'tech.units:indriya:2.1.2'
 
 	// JTS Topology Suite for GeoPositions, License: EPL 1.0 / EDL 1.0
 	compile 'org.locationtech.jts:jts-core:1.16.1'
--- a/src/main/java/edu/ie3/datamodel/models/result/system/SystemParticipantResult.java
+++ b/src/main/java/edu/ie3/datamodel/models/result/system/SystemParticipantResult.java
@@ -15,12 +15,10 @@ import tech.units.indriya.ComparableQuantity;
 /** Abstract class that holds values common to all other result entities */
 public abstract class SystemParticipantResult extends ResultEntity {
 
-  /**
-   * @param p active power output normally provided in MW
-   * @param q reactive power output normally provided in MVAr
-   */
+  /** active power output normally provided in MW */
   private ComparableQuantity<Power> p;
 
+  /** reactive power output normally provided in MVAr */
   private ComparableQuantity<Power> q;
 
   /**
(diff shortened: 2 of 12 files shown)


build.gradle
```diff
@@ -54,7 +54,7 @@ dependencies {
 	//	 ie³ power system utils
 	compile 'com.github.ie3-institute:PowerSystemUtils:1.5+'
 
-	implementation 'tech.units:indriya:2.0.4'
+	implementation 'tech.units:indriya:2.1.2'
```
ckittl, line 57: Why is this change needed? I'm a bit anxious about updating the units API, as it has brought a lot of bugs and side effects in the past.

src/main/java/edu/ie3/datamodel/models/result/system/SystemParticipantResult.java
```diff
@@ -15,12 +15,10 @@
 /** Abstract class that holds values common to all other result entities */
 public abstract class SystemParticipantResult extends ResultEntity {
 
-  /**
-   * @param p active power output normally provided in MW
-   * @param q reactive power output normally provided in MVAr
-   */
+  /** active power output normally provided in MW */
```
ckittl, line 18: I think, it's okay, but it might be an idea to link `StandardUnits#ACTIVE_POWER_OUTPUT` (or however this static field is called)?!

src/main/java/edu/ie3/datamodel/utils/ContainerUtils.java
```diff
@@ -821,7 +821,8 @@ public static SubGridContainer withTrafoNodeAsSlack(final SubGridContainer subGr
             newNodes,
             subGridContainer.getRawGrid().getLines(),
             newTrafos2w,
-            newTrafos3wToInternalNode.keySet(),
+            // HashSet$KeySet is not serializable, thus create new set
```
ckittl, line 824: Great, that you have added this comment!

ckittl (approved):
Please add your work to the CHANGELOG.md.

Generally, the changes look good to me. But please answer my question regarding the update of units API and consider thoroughly, if we need that update. Moreover, #283 feels to be an issue, that might stay open after this PR? So maybe link it in another way.

build.gradle
sebastian-peter: Indriya in versin 2.0.4 does not serialize some quantities properly. There's an error thrown because a transient field is not re-initialized. I picked 2.1.2 because it was the newest version and it fixed the issue, but if you like I can try to find an earlier version that works as well.

src/main/java/edu/ie3/datamodel/models/result/system/SystemParticipantResult.java
sebastian-peter: I just did this because the compiler was complaining about the @param in JavaDoc of a field. I just fixed the JavaDoc to make it work. Not quite sure what addition you have in mind here.

sebastian-peter:
> Moreover, #283 feels to be an issue, that might stay open after this PR? So maybe link it in another way.

At first I opened a general issue regarding distributed computation, but as our plan regarding merges to main changed, I thought it would be wise to close this issue and open new ones, if there's the need.

PR #329, opened 2021-04-16
Adapt column scheme of ICON weather data

ckittl:
Closes #267. Close #321 beforehand.

Adapting the ICON weather data scheme + integration tests with various technologies + minor improvements.

PR #331, opened 2021-04-16
Improve handling of time stamp pattern in factories

### LINKED ISSUE #330: Consistency check for date time pattern if using InfluxDB
When querying InfluxDB, results contain a time stamp in [RFC 3339](https://tools.ietf.org/html/rfc3339) format. (cf. [InfluxDB doc](https://docs.influxdata.com/influxdb/v2.0/reference/glossary/#rfc3339-timestamp)). However, currently, the factories being used for parsing, may rely on a different kind of format.

A similar issue arises with other sources, that might query based upon a notion of time. I would suggest to make the factories' time stamp pattern accessible from outside, so that either consistency checks are possible or one can use the factory's pattern for other purposes.

ckittl:
Closes #330. Take care of #329 beforehand.


staudtMarius:
@danielfeismann, @sebastian-peter I think this PR and the related issue can be closed, since we removed the time pattern input from the factories.

PR #395, opened 2021-07-08
Replace LogManager with LogFactory

### LINKED ISSUE #394: Replace LogManager with LoggerFactory
Some classes invalidly instantiate logger with `org.apache.logging.log4j.LogManager` instead of `org.slf4j.LoggerFactory` which contradicts the idea of slf4j as logging facade. 

We need to fix this to avoid breaking implementations that make use of other logging frameworks e.g. logback.

johanneshiry:
Resolves #394


johanneshiry:
@reviewer: Please take a look if it is really necessary to fix the QualityGate as only error logging is affected

ckittl (approved):

PR #407, opened 2021-08-16
Introducing EvcsLocationType

### LINKED ISSUE #406: Enhance EvcsInput/EvcsInputFactory to include EvcsLocationType
EvcsLocationType should be implemented in this project as an enum and read exist as a field in EvcsInput csv

Maybe there needs to be a discussion about what the location types should be

sebastian-peter:
Resolves #406

### DIFF
--- a/src/main/java/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtils.java
+++ b/src/main/java/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtils.java
@@ -1,0 +1,30 @@
+/*
+ * © 2021. TU Dortmund University,
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.models.input.system.type.evcslocation;
+
+import edu.ie3.datamodel.exceptions.ParsingException;
+import java.util.HashMap;
+
+public class EvcsLocationTypeUtils {
+
+  private static final HashMap<String, EvcsLocationType> nameToType = initMap();
+
+  private static HashMap<String, EvcsLocationType> initMap() {
+    final HashMap<String, EvcsLocationType> map = new HashMap<>(EvcsLocationType.values().length);
+    for (EvcsLocationType type : EvcsLocationType.values()) map.put(type.name(), type);
+
+    return map;
+  }
+
+  private EvcsLocationTypeUtils() {
+    throw new IllegalStateException("This is a factory class. Don't try to instantiate it.");
+  }
+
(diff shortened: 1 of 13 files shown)


src/main/java/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtils.java
```diff
@@ -0,0 +1,30 @@
...
+
+    return map;
+  }
+
+  private EvcsLocationTypeUtils() {
+    throw new IllegalStateException("This is a factory class. Don't try to instantiate it.");
+  }
+
+  public static EvcsLocationType parse(String parsableString) throws ParsingException {
+    if (nameToType.containsKey(parsableString)) return nameToType.get(parsableString);
```
ckittl, line 27: I would wish for a bit less restrictive way of parsing, e.g. comparing with neglecting the case (cf. `BdewLoadProfile`). And we should consider, which case (camel, snake, ...) we actually expect. Any suggestion?

src/test/groovy/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtilsTest.groovy
```diff
@@ -0,0 +1,56 @@
...
+
+	def "The EvcsLocationTypeUtils should throw exceptions as expected when invalid evcs location type string is provided"() {
+		when:
+		parse(invalidString)
+
+		then:
+		ParsingException ex = thrown()
+		ex.message == expectedExceptionMsg
+
+		where:
```
ckittl, line 51: I think, data driven testing isn't necessary in this case.

ckittl (changes requested):
Please address the comments and adapt the Documentation (readthedocs). Especially, it is important to denote the range of permitted values and how to enter them.

src/main/java/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtils.java
sebastian-peter: I wrote the type names in upper & snake case because that's the Java convention regarding enums. Of course other styles are possible, we have to expect SonarLint warnings then though.

src/test/groovy/edu/ie3/datamodel/models/input/system/type/evcslocation/EvcsLocationTypeUtilsTest.groovy
sebastian-peter: Are you refering to the checking of the error message, or the whole test altogether?

ckittl (approved):

PR #505, opened 2022-01-17
No/#405 add olm model

### LINKED ISSUE #405: Add OLM model
Add new OLM model to support different implementations of olm

noffermann:
resolves #405


docs/readthedocs/models/input/OLM.rst
```diff
@@ -0,0 +1,45 @@
...
+""""""""""
+
++--------------+---------+---------------------------------------------+
+| Attribute    | Unit    | Remarks                                     |
++==============+=========+=============================================+
+| uuid         | --      |                                             |
++--------------+---------+---------------------------------------------+
+| id           | --      | Human readable identifier                   |
++--------------+---------+---------------------------------------------+
+| T_s          |  °C     | max. line temperature                       |
```
ckittl, line 20: Throughout the PowerSystemDataModel our variable naming shifted a bit towards expressiveness. It would be good, if you could consider some changes in the naming. E.g. `temperatureMax`.  Additionally, we have used camel case in the other documentation pages. So, it would be good, if we could stick with the same concept in all places.

Moreover, we have defined some [StandardUnits](https://github.com/ie3-institute/PowerSystemDataModel/blob/dev/src/main/java/edu/ie3/datamodel/models/StandardUnits.java). Maybe you could consider to switch the units to those.

docs/readthedocs/models/input/OLM.rst
```diff
@@ -0,0 +1,45 @@
...
++--------------+---------+---------------------------------------------+
+| alpha_line   | 1/K     | Coefficient of thermal expansion            |
++--------------+---------+---------------------------------------------+
+| layer        | --      | Number of layers per line                   |
++--------------+---------+---------------------------------------------+
+| sq           | mm^2    | cross section                               |
++--------------+---------+---------------------------------------------+
+| D_0          | m       | Leiterdurchmesser Bewehrung                 |
```
ckittl, line 28: As we want to target a wide audience, please use English for the documentation.

docs/readthedocs/models/input/OLM.rst
```diff
@@ -0,0 +1,45 @@
...
++--------------+---------+---------------------------------------------+
+| R_dc20       | Ohm/m   | Specific DC resistance at 20°C              |
++--------------+---------+---------------------------------------------+
+| Buendelanzahl| --      |                                             |
```
ckittl, line 32: Same here.

docs/readthedocs/models/input/grid/line.rst
```diff
@@ -59,11 +59,6 @@ Entity Model
 | geoPosition       | --   | | Line string of geographical locations describing the |
 |                   |      | | position of the line                                 |
 +-------------------+------+--------------------------------------------------------+
-| olmCharacteristic | --   | | Characteristic of possible overhead line monitoring  |
```
ckittl, line 62: Within the line model, we need some reference to the OLM model. That is, that I'd like to keep this field and possibly rename it to `olmModel` or so. Moreover, it would also be desired, that we can have simple models as well. That said, would you mind proposing a second olm model (in the way with the above file) to keep the functionality of this simple model?

ckittl (changes requested):
Thanks a lot for your draft!

If we can get convergence on this documentation, we can start implementing the model (wouldn't like to merge a documentation, that does not reflect the actual state of code). Currently we are about and on our way to have a release with breaking changes in data scheme. I will have a discussion with the other team mates, if we can't / wouldn't like to include your changes in that release as well.

PR #530, opened 2022-02-11
To/#529 get nearest coordinates

### LINKED ISSUE #529: Adapt `IdCoordinateSource.getNearestCoordinates` to explicitly use all coordinates

t-ober:
Resolves #529


src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -66,17 +66,15 @@
    *
    * @param coordinate the coordinate to look up the nearest neighbours for
    * @param n how many neighbours to look up
-   * @param allCoordinates the collection of points, ideally containing all available coordinates
+   * @param coordinates the collection of points
```
t-ober, line 69: did change the documentation because ideally you would only look through a minimal set of coordinates that include the nearest coordinates

johanneshiry (changes requested):
Please fix as discussed + also adapt the issue title and description. Thank you! :)

johanneshiry (approved):

PR #550, opened 2022-02-28
use instanceof instead of getClass for equals methods in input models

### LINKED ISSUE #354: Fix equals methods in input models
Currently, the Object#equals method in almost every class us using `getClass` instead of `instanceof`. This breaks substitutability of subclasses. 

There are two possible ways to deal with this:
- either make every class that implements equals that way final
- use instanceof instead of getClass

danielfeismann:
Resolves #354


danielfeismann:
!test

PR #552, opened 2022-03-01
Exclude spock milesone versions

sebastian-peter:
`M` versions are milestone versions (somewhere between snapshot and release candidate), while we should use release versions. Thus, exclude milestone versions.

Also included are probable future versions (`2.2-M2`) which might come up at some point.


johanneshiry (approved):
LGTM, thank you!

PR #556, opened 2022-03-02
Parse with default

### LINKED ISSUE #555: Issue with parsing of time stamp in IconTimeBasedWeatherValueFactory
The `buildModel` function does not parse time pattern with defaults which leads to `DateTimeParseException`

t-ober:
Resolves #555


johanneshiry (changes requested):
This breaks logic and is definitely not what you want. 

Take a look at the constructor: you can already pass in any instance of TimeUtil (either with defaults or not). With your change this is not possible anymore and the class field `timeUtil` becomes useless. Cannot determine where your bug in the issue comes from bc of lack of details, but I would _assume_ the issue is an invalid instantiation of the factory before using it!

PR #321, opened 2021-04-06
Change scheme of COSMO weather data

ckittl:
Resolves parts of #267.

~Close #319 before!~

Includes:
- Adaption of column scheme for COSMO model
- Renaming of all classes which had `Psdm` in it to `Cosmo`
- Consolidate weather test files
- Offer possibility to have different naming strategies

Issue #267, opened 2020-12-17
Fix PsdmTimeBasedWeatherValueFactory


johanneshiry:
The column names of the `PsdmTimeBasedWeatherValueFactory` are currently invalid.

- time -> time 
- coordinate -> **coordinateid**
- diffuseirradiation -> **diffuseirradiance**
- directirradiation -> **directirradiance**
- winddirection -> winddirection
- windvelocity -> windvelocity
- temperature -> temperature

Furthermore the `datum` column of the`IconTimeBasedWeatherValueFactory` might be renamed to `time` as well. (considering #175)

Furthermore, `Psdm` should be renamed to `Cosmo`.

ckittl:
As of discussion today: The factory does not reflect the _actual_ scheme in the weather database. Therefore, either the factory has to be altered or at the same time the scheme of the data base.

An argument for changing the data scheme is, that InfluxDB requires the time information to be called `time`. On the other hand, not changing the scheme, would not lead to breaking other applications.

Decision:
- [ ] Alter the scheme for the COSMO model within the data base, as this kind of is legacy data as well.
- [ ] Rename `PsdmTimeBaseWeatherValueFactory` to `CosmoTimeBasedWeatherValueFactory`
- [ ] Alter the scheme of ICON weather data as well (`datum` -> `time`, `coordinate` -> `coordinateid`) -- change the column name here as well

ckittl:
Even more issues than expected...

- [ ] `CouchbaseWeatherSource` takes name of coordinate column from constructor, `InfluxDbWeatherSource` from `TimeBasedWeatherValueFactory`-implementation for queries -> Resolve this inhomogeneity
- [ ] Field names in Couchbase (the actual source) are plain and not snake case as in other sources -> can it be harmonized?

ckittl:
Breaking down the above discussion to actual tasks:

- [x] Adaption of COSMO column scheme in code
  - targeted in: #321 
- [ ] Adapt COMSO column scheme in all existing sources (files, database, ...)
- [x] Adapt ICON column scheme in code
  - targeted in: #329
- [ ] Adapt ICON column scheme in all existing sources (files, database, ...)
- [ ] `SqlWeatherSource` determines column name by querying the meta scheme
  - [ ] Check where this is actually used
  - [ ] Try to get rid of that implementation and use fixed column name convention instead
- [ ] Change Couchbase column names from flat to snake case
[...]

PR #558, opened 2022-03-03
Adapt to changed scheme of COSMO weather data

sebastian-peter:
Adaption of #321 by @ckittl without introducing `NamingConvention` to parts of the sources ([more here](https://github.com/ie3-institute/PowerSystemDataModel/pull/321#discussion_r805770366))

Resolves parts of #267.

Includes:
- Adaption of column scheme for COSMO model
- Renaming of all classes which had `Psdm` in it to `Cosmo`
- Consolidate weather test files


src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -30,7 +30,11 @@
   private static final String WHERE = " where ";
   private static final String MEASUREMENT_NAME_WEATHER = "weather";
   private static final int MILLI_TO_NANO_FACTOR = 1000000;
-  private final String coordinateIdColumnName;
+
+  /* Final name of the coordinate id field for use in factories */
+  private final String coordinateIdFieldName;
+  /* Final name of the column within the database */
+  private static final String coordinateIdColumnName = "coordinate_id";
```
t-ober: reminder to address sonarqube issue

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -30,7 +30,11 @@
   private static final String WHERE = " where ";
   private static final String MEASUREMENT_NAME_WEATHER = "weather";
   private static final int MILLI_TO_NANO_FACTOR = 1000000;
-  private final String coordinateIdColumnName;
+
+  /* Final name of the coordinate id field for use in factories */
+  private final String coordinateIdFieldName;
```
t-ober, line 36: why do we track that here as an attribute explicitly. If it's just the value of the `WeatherFactory` we can just access it via the factory

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -50,7 +54,7 @@ public InfluxDbWeatherSource(
     this.connector = connector;
     this.coordinateSource = coordinateSource;
     this.weatherValueFactory = weatherValueFactory;
-    this.coordinateIdColumnName = weatherValueFactory.getCoordinateIdFieldString();
+    this.coordinateIdFieldName = weatherValueFactory.getCoordinateIdFieldString();
```
t-ober, line 57: why do we track that here as an attribute explicitly. If it's just the value of the WeatherFactory we can just access it via the factory

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -152,24 +156,25 @@ public InfluxDbWeatherSource(
     return measurementsMap.get(MEASUREMENT_NAME_WEATHER).stream()
```
t-ober, line 156: something weird is going on in the docs of the function of `optTimeBasedValueStrem` on line 149 - would be nice if we can fix that here as well

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -152,24 +156,25 @@ public InfluxDbWeatherSource(
...
-              Map<String, String> camelCaseFields =
+              /* The factory expects flat case id's for fields -> Convert the keys */
+              Map<String, String> flatCaseFields =
                   fieldToValue.entrySet().stream()
                       .collect(
                           Collectors.toMap(
-                              entry -> StringUtils.snakeCaseToCamelCase(entry.getKey()),
+                              entry ->
+                                  StringUtils.snakeCaseToCamelCase(entry.getKey()).toLowerCase(),
                               Map.Entry::getValue));
```
t-ober, line 166: Do you think it might make sense to extract this to a different function? Would reduce complexity of this function.

t-ober (changes requested):
Looks good - just minor stuff

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
sebastian-peter: That's true

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
sebastian-peter: With the new functionalities in PSU, I can imagine a universal converion tool for naming formats: Convert from an arbitrary format to another using `NamingConvention`. I created a new issue here: https://github.com/ie3-institute/PowerSystemUtils/issues/235

I'd suggest to solve issues like this using this new functionality (in the future), which would also make parts of `StringUtils` obsolete.

t-ober (approved):

PR #564, opened 2022-03-08
Write csv entries in correct order

### LINKED ISSUE #434: Properly write out data given by entity field mapping
https://github.com/ie3-institute/PowerSystemDataModel/blob/124d2fcacfdbc2c3cea6c91631f98db037419780/src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java#L67-L86

Current implementation of write method exhibits wrong behavior. It assumes, that the map entries are given in the same order as the headline elements. However, this is not always the case and especially not the behavior, that one would expect. This possibly leads to data errors, where you can't spot a swapped column (because both are numerical and of same range).

ckittl:
Resolves #434

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
+++ b/src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
@@ -27,10 +27,11 @@ public class BufferedCsvWriter extends BufferedWriter {
       "Direct appending is prohibited. Use write instead.";
 
   /**
-   * Build a new CsvBufferedWriter
+   * Build a new CsvBufferedWriter. The order of headline elements given in this constructor defines
+   * the order of columns in file
    *
    * @param filePath String representation of the full path to the target file
-   * @param headLineElements Elements of the csv head line
+   * @param headLineElements Elements of the csv headline
    * @param csvSep csv separator char
    * @param append true to append to an existing file, false to overwrite an existing file (if any),
    *     if no file exists, a new one will be created in both cases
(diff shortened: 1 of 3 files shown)


src/test/groovy/edu/ie3/datamodel/io/csv/BufferedCsvWriterTest.groovy
```diff
@@ -78,4 +79,33 @@ class BufferedCsvWriterTest extends Specification {
...
+	def "The buffered csv writer writes out content in the order specified by the headline elements"() {
+		given:
+		def targetFile = FilenameUtils.concat(tmpDirectory.toString(), "order_test.csv")
+		def writer = new BufferedCsvWriter(targetFile, ["c", "b", "a"] as String[], ",", false)
+		writer.writeFileHeader()
+		def content = [
+				"c": "z",
+				"a": "x",
+				"b": "y"
+		]
```
johanneshiry, line 92: Maybe slightly more expressive headline-value-pairs may improve readability of this? Quite hard to check the test  lines in 108 & 109

src/main/java/edu/ie3/datamodel/io/csv/BufferedCsvWriter.java
```diff
@@ -73,15 +75,16 @@ public BufferedCsvWriter(String baseFolder, CsvFileDefinition fileDefinition, bo
...
         || !entityFieldData.keySet().containsAll(Arrays.asList(headLineElements)))
       throw new SinkException(
           "The provided data does not meet the pre-defined head line elements '"
               + String.join(",", headLineElements)
               + "'.");
 
-    String[] entries = entityFieldData.values().toArray(new String[0]);
+    String[] entries =
+        Arrays.stream(headLineElements).map(entityFieldData::get).toArray(String[]::new);
     writeOneLine(entries);
```
johanneshiry, line 88: Here, a stream terminal operation (toArray) is used. You may consider adapting the writeOneLine to take a Stream as input parameter. This would allow for out of the box appending the elements of the stream avoiding the overhead of array allocation. But this is a very minor thing that came into my mind when reading this line ...

johanneshiry (dismissed):
Good catch, thank you! Approving now but you may take into account my comments though if you want to :)

danielfeismann (approved):
Looks good, thank you :)

danielfeismann (approved):

PR #583, opened 2022-04-27
apply rfc compliance to csv sink

### LINKED ISSUE #582: CSVSink does not generate RFC compliant csv files
PR [33](https://github.com/ie3-institute/PowerSystemUtils/pull/33) introduced RFC compliancy for csv sinks and sources but the actual remapping was not applied to the sink.

t-ober:
Resolves #582

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
+++ b/src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
@@ -324,18 +324,19 @@ public class CsvFileSink implements InputDataSink, OutputDataSink {
     LinkedHashMap<String, String> entityFieldData;
     try {
       entityFieldData =
-          processorProvider
-              .handleEntity(entity)
-              .orElseThrow(
-                  () ->
-                      new SinkException(
-                          "Cannot persist entity of type '"
-                              + entity.getClass().getSimpleName()
-                              + "'. This sink can only process the following entities: ["
-                              + processorProvider.getRegisteredClasses().stream()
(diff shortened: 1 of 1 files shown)


src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -324,18 +324,19 @@ private <C extends UniqueEntity> void write(C entity) {
...
-                          "Cannot persist entity of type '"
-                              + entity.getClass().getSimpleName()
-                              + "'. This sink can only process the following entities: ["
-                              + processorProvider.getRegisteredClasses().stream()
-                                  .map(Class::getSimpleName)
-                                  .collect(Collectors.joining(","))
-                              + "]"));
+          csvEntityFieldData(
+              processorProvider
+                  .handleEntity(entity)
```
sebastian-peter, line 329: I'd rather apply `csvEntityFieldData` via `map` here so that indentation does not get too crazy. So it'd be
```
.handleEntity(entity)
.map(csvEntityFieldData)
.orElseThrow(...
```
or similar.

sebastian-peter (changes requested):
Just a tiny comment here...

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
t-ober: Fair point !

sebastian-peter:
Looks good to me. In the future, we should provide an integration test to prevent such issues from happening though -> #586

PR #608, opened 2022-06-03
To/#607 profile parse method

### LINKED ISSUE #607: Implement separate parse method for different load profile classes

t-ober:
resolves #607


sebastian-peter (approved):
I could imagine an implementation that does not rely on casting, but I'm also fine with this

sebastian-peter:
Let me work out this concept I have in mind...

PR #609, opened 2022-06-03
Sp/#607 profile parse method

### LINKED ISSUE #607: Implement separate parse method for different load profile classes

sebastian-peter:
Resolves #607 

Builds on contributions of #608 by @t-ober 

At some point, when the number of load profiles grows, we should consider caching the profiles returned by `getAllProfiles()`. For now, I think it's fine the way it is.


t-ober (approved):
Makes sense - thanks :)

PR #611, opened 2022-06-04
To/#610 extend temp lps

### LINKED ISSUE #610: Extend `TemperatureDependantLoadProfiles` with reference temperature and minimal temperature

t-ober:
resolves #610

### DIFF
--- a/src/main/java/edu/ie3/datamodel/models/profile/NbwTemperatureDependantLoadProfile.java
+++ b/src/main/java/edu/ie3/datamodel/models/profile/NbwTemperatureDependantLoadProfile.java
@@ -7,14 +7,12 @@ package edu.ie3.datamodel.models.profile;
 
 import edu.ie3.datamodel.exceptions.ParsingException;
 import edu.ie3.datamodel.models.StandardUnits;
+import javax.measure.quantity.Temperature;
 import tech.units.indriya.ComparableQuantity;
 import tech.units.indriya.quantity.Quantities;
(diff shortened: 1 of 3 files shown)


t-ober:
@sebastian-peter I'm not really satisfied with this implementation as the `referenceTemperature` and `minTemperature` should be static. But sine you can not declare abstract static fields nor methods I don't find a way to declare them in the interface but make the subclasses set the specific value. I'm probably missing the right way to do it. Maybe you can help me out once more 👼🏼

sebastian-peter:
@t-ober since the (power) values are fully handled by SIMONA and the within PSDM we (up until now) only provide keys and inputs, wouldn't it make sense to provide this info in SIMONA instead?

t-ober:
> @t-ober since the (power) values are fully handled by SIMONA and the within PSDM we (up until now) only provide keys and inputs, wouldn't it make sense to provide this info in SIMONA instead?

@sebastian-peter I thought about this too but I that information is part of the actual input as the calculation gets parameterized by it and we need to make sure that this gets supplied for every subclass of temperature dependent profiles as the values differ between different profile of different grid providers. So if you would put them in SIMONA for adding a different profile set e.g. `WestnetzTemperatureDependentLoadProfiles` we would need to make a new datastructure in SIMONA setting up the additional parameters. All in all I think this should be part of the PSDM data model as it specifically is _input_ for the calculation later on.

PR #694, opened 2022-10-27
SQL source for weather coordinates using PostGIS

### LINKED ISSUE #689: Improving implementation of IdCoordinateSource
This issue is realated to the following issue:  
- https://github.com/ie3-institute/simona/issues/145

This issue aims to improve the implementation of ``IdCoordinateSource`` to find the nearest coordinates for a point much more efficiently.

staudtMarius:
Resolves #689


src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -0,0 +1,315 @@
...
+
+  /**
+   * Method for evaluating the found points and returning the n corner points of the bounding box.
+   *
+   * @param coordinate at the center of the bounding box
+   * @param distances list of found points with their distances
+   * @param numberOfPoints that should be returned
+   * @return list of distances
+   */
+  private List<CoordinateDistance> checkForBoundingBox(
```
sebastian-peter, line 183: Could you find a more fitting name for this method? Maybe something like `extractBoundingBox` or `restrictToBoundingBox`...

sebastian-peter (changes requested):
This is my first quick review. Please hit me up if you have any questions.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -0,0 +1,315 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.locationtech.jts.geom.Point;
+
+/** SQL source for coordinate data */
+public class SqlIdCoordinateSource extends SqlDataSource<CoordinateValue>
+    implements IdCoordinateSource {
+  private static final String WHERE = " WHERE ";
+  private final IdCoordinateFactory factory;
+  private final double maxDistance;
+  private static final double earthRadius = 6378137.0;
```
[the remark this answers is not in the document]
staudtMarius: Unfortunatlely, it is saved as a private variable. Should I change this in PSU?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -0,0 +1,315 @@
...
+              ps.setDouble(1, latitude);
+              ps.setDouble(2, longitude);
+            });
+
+    return Optional.of(values.get(0).id);
+  }
+
+  @Override
+  public Collection<Point> getAllCoordinates() {
+    List<CoordinateValue> values = executeQuery(basicQuery, ps -> ps.setString(1, ";"));
```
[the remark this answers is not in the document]
staudtMarius: I need the parameter to return all available points.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -75,6 +74,94 @@
         GeoUtils.calcOrderedCoordinateDistances(
             coordinate,
             (coordinates != null && !coordinates.isEmpty()) ? coordinates : getAllCoordinates());
-    return sortedDistances.stream().limit(n).toList();
+    return restrictToBoundingBoxWithSetNumberOfCorner(coordinate, sortedDistances, n);
+  }
+
+  /**
+   * Method to turn a distance into a latitude and longitude deltas. The methode can be found here:
+   * <a href="https://math.stackexchange.com/questions/474602/reverse-use-of-haversine-formula">
```
sebastian-peter, line 82: Link is missing a closing tag

sebastian-peter (changes requested):
I wrote these comments before I thought about exploring PostGIS, so it might make sense to explore that before looking through some comments

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -75,6 +74,94 @@
...
+    // by the earth's radius to get a y-delta in radians
+    double deltaY = maxDistance / earthRadius;
+
+    // because the spacing between x-degrees change between the equator
+    // and the poles, we need to calculate the x-delta using the inverse
+    // haversine formula
+    double sinus = Math.sin(deltaY / 2);
+    double squaredSinus = sinus * sinus;
+    double cosine = Math.cos(Math.toRadians(coordinate.getY()));
+    double squaredCosine = cosine * cosine;
```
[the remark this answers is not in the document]
staudtMarius: This method will be moved to PSU. All requested changes will be implemented there. 
- https://github.com/ie3-institute/PowerSystemUtils/pull/321

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
```diff
@@ -75,6 +74,94 @@
...
+  }
+
+  /**
+   * Method to turn a distance into a latitude and longitude deltas. The methode can be found here:
+   * <a href="https://math.stackexchange.com/questions/474602/reverse-use-of-haversine-formula">
+   *
+   * @param coordinate the coordinate at the center of the bounding box.
+   * @return x- and y-delta in degree
+   */
+  default double[] calculateXYDelta(Point coordinate, double maxDistance) {
```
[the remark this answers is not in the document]
staudtMarius: See above.

src/main/java/edu/ie3/datamodel/io/source/IdCoordinateSource.java
staudtMarius: See https://github.com/ie3-institute/PowerSystemUtils/pull/321

src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -0,0 +1,215 @@
...
+              ps.setArray(1, sqlArray);
+            });
+
+    List<Point> points = new ArrayList<>();
+
+    for (CoordinateValue value : values) {
+      points.add(value.coordinate);
+    }
+
+    return points;
```
sebastian-peter, line 107: This can be simplified. Please check in this file for further similar occurrences.

```suggestion
    return values.stream().map(value -> value.coordinate).toList();
```

src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    return Optional.of(new CoordinateValue(id, point));
+  }
+
+  @Override
+  public Optional<Point> getCoordinate(int id) {
+    List<CoordinateValue> values = executeQuery(queryForPoint, ps -> ps.setInt(1, id));
+
+    return Optional.of(values.get(0).coordinate);
```
sebastian-peter, line 86: What happens if no row for given id is found? With this implementation, there will be an exception thrown, but you actually rather want an empty Optional to be returned. Please check for further occurrences.

sebastian-peter (changes requested):
Thanks, looks good overall. Found some ways to further improve your implementation.

src/main/java/edu/ie3/datamodel/io/factory/timeseries/CosmoIdCoordinateFactory.java
```diff
@@ -51,4 +52,9 @@ public String getLatField() {
   public String getLonField() {
     return LONG_GEO;
   }
+
+  @Override
+  public String getPointField() {
+    return Point;
+  }
```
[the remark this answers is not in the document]
t-ober: It seems like we only have the different coordinate sources to be able to parse and differentiate the `Cosmo` from the `Icon` coordinates. They are in so far different as they represent different coordinate grids which belong to the respective DWD model. `Icon` being their more international weather model and `Cosmo` a more regional one. Therefore we definitely need to make sure that we don't interchange them. Internally after initialization we only care about the `lat` and `lon` values and the id mapping. If we rather directly want a `Point` representation in the table which it seems like we do I do not see any reason to keep the `lat` and `lon` values as they only duplicate information.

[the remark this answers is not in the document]
sebastian-peter: That's the direction that I thought of. So it'd make sense to have a third type for coordinate tables that use geometries, right? And then I'd make this type mandatory for SQL coordinates, unless we also want the (slower) implementation using lat/lon columns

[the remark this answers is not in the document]
sebastian-peter: @staudtMarius After the discussion I had with @t-ober we came to the conclusion that creating a third coordinate type besides COSMO and ICON coordinates would be best. So that'd be a CoordinateFactory that only uses `id` and `point` (maybe rename it to `coordinate`) as fields.

sebastian-peter (changes requested):
Thanks for your contributions!

Besides my (smaller) comments here, while I was looking through the existing and your new code I realized the rationale behind what is happening here. Since there's no documentation for coordinate sources yet, it'd be great if you could start this in readthedocs. I started with the text below while trying to understand the code (it's just a rough starting point though).

For a given set of coordinates, the `n` closest points from the four quadrants, plus more close points if `n > 4`. The input set of coordinates is either the coordinates found within the bounding box, or `n` nearest neighbors if no coordinates exist within the box.

sebastian-peter (changes requested):
Ok I think it'd be great if the coordinates documentation would be divided into the parts

a) below [models/input](https://powersystemdatamodel.readthedocs.io/en/latest/models/models.html#input) with general information of what coordinate sources do. This could be below a `Weather` headline, where you do not need to provide details about weather data, but only a subsection about coordinates

b) specific hints on coordinate source types, i.e. [CSV](https://powersystemdatamodel.readthedocs.io/en/latest/io/csvfiles.html) and SQL (needs to be created), within the I/O section.

Please let me know if we should go more into detail regarding the outlines of these sections.

sebastian-peter (changes requested):
I've skimmed through the changes and it looks quite good so far, I'll pay more attention to details later though.

One thing that caught my eye is that readthedocs now fails, I think you have to update myst-parser now that sphinx has been updated.
Also, if I execute sphinx-build locally (with a fixed configuration), there's a bunch of warnings - could you fix them? I took care of all warnings that existed before in `dev`, so I think it should be possible to eliminate the new ones :)

docs/readthedocs/io/sql.md
```diff
@@ -0,0 +1,7 @@
+# Sql
```
sebastian-peter, line 1: Please write `SQL` in all-caps

docs/readthedocs/io/csvfiles.md
```diff
@@ -0,0 +1,211 @@
+# csv files
```
sebastian-peter, line 1: CSV

docs/readthedocs/models/input/additionaldata/timeseries.md
```diff
@@ -0,0 +1,30 @@
+# Time Series
+Time series are meant to represent a timely ordered series of values.
+Those can either be electrical or non-electrical depending on what one may need for power system simulations.
+Our time series models are divided into two subtypes:
+
+**Individual Time Series** <br>
```
sebastian-peter, line 6: Please use Secondary heading here: `##`

sebastian-peter (changes requested):
Haven't checked everything, but here's some things to do again :)

sebastian-peter:
!test

docs/readthedocs/io/source/idcoordinatesource.md
```diff
@@ -0,0 +1,84 @@
...
+
+The IdCoordinateSource contains a method for retrieving the id of a given coordinate.
+
+    Optional<Integer> getId(Point coordinate)
+
+This method is used to return the id of a given coordinate. If no id is found for the given
+coordinate, an empty optional is returned.
+
+
+## Method for calculating distances:
```
sebastian-peter, line 50: "Methods for retrieving near coordinates"

docs/readthedocs/io/source/idcoordinatesource.md
```diff
@@ -0,0 +1,84 @@
...
+The IdCoordinateSource also contains methods for calculation the distances og a given coordinate
```
sebastian-peter, line 51: typo

docs/readthedocs/io/source/idcoordinatesource.md
```diff
@@ -0,0 +1,84 @@
...
+    Optional<Integer> getId(Point coordinate)
+
+This method is used to return the id of a given coordinate. If no id is found for the given
+coordinate, an empty optional is returned.
+
+
+## Method for calculating distances:
+The IdCoordinateSource also contains methods for calculation the distances og a given coordinate 
+to a set of coordinates. All the following methods will return the closest n coordinates with their 
+distances.
```
sebastian-peter, line 53: Update, please

docs/readthedocs/io/source/idcoordinatesource.md
```diff
@@ -0,0 +1,84 @@
...
+the method will return the closest n coordinates. If the number of distances < n, this method will
+return less than n coordinates.
+
+
+## Finding and returning the closest corner coordinates:
+In most cases we need four corner coordinates for our given coordinate. Therefor the 
+IdCoordinateSource contains a method that will use the calculated distances to find the closest 
+corner coordinates for the given coordinate.
+
+    List<CoordinateDistance> restrictToBoundingBoxWithSetNumberOfCorner(
```
sebastian-peter, line 75: Update, please

docs/readthedocs/io/csvfiles.md
```diff
@@ -0,0 +1,211 @@
...
+    sink.persistJointGridContainer(grid);
+
+The sink takes a collection of model suitable for serialization and handles the rest (e.g. unboxing of nested models)
+on its own.
+But caveat: As the (csv) writers are implemented in a concurrent, non-blocking way, duplicates of nested models could
+occur.
+
+## Compression and extraction of files
+
+We consider either regular directories or compressed <a href="https://en.wikipedia.org/wiki/Tar_(computing)">tarball archives</a>
```
sebastian-peter, line 209: Could you also convert the link to md syntax?

src/main/java/edu/ie3/datamodel/io/factory/timeseries/SqlCoordinateFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  public String getLatField() {
+    return null;
+  }
+
+  @Override
+  public String getLonField() {
+    return null;
+  }
```
sebastian-peter, line 59: Is it possible to throw exceptions here instead of returning null?

sebastian-peter (changes requested):

staudtMarius:
!test

docs/readthedocs/conf.py
```diff
@@ -35,10 +35,14 @@
...
+    'sphinx.ext.autodoc',
+    'sphinx.ext.intersphinx',
+    'myst_parser'
 ]
 
+myst_enable_extensions = ["dollarmath", "amsmath"]
+myst_heading_anchors = 4
+
 # Prefix all autogenerated labels wit the document to get unique labels (e.g. `index:Hello`)
 autosectionlabel_prefix_document = True
```
sebastian-peter, line 47: This can be removed now that `autosectionlabel is gone`

docs/readthedocs/conf.py
```diff
@@ -35,10 +35,14 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'recommonmark',
-    'sphinx.ext.autosectionlabel'
+    'sphinx.ext.autodoc',
```
sebastian-peter, line 38: Do we need this somewhere?

docs/readthedocs/conf.py
```diff
@@ -18,12 +18,12 @@
...
 project = 'PowerSystemDataModel'
-copyright = u'2020. TU Dortmund University, Institute of Energy Systems, Energy Efficiency and Energy Economics, Research group Distribution grid planning and operation '
+copyright = u'2023. TU Dortmund University, Institute of Energy Systems, Energy Efficiency and Energy Economics, Research group Distribution grid planning and operation '
 author = 'Johannes Hiry, Debopama Sen Sarma, Chris Kittl'
 
 # The full version, including alpha/beta/rc tags
-version = '1.0'
-release = '1.0.1-SNAPSHOT'
+version = '2.1.0'
+release = '3.0-SNAPSHOT'
```
sebastian-peter, line 26: Please check out the [documentation for version and release](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-version)

docs/readthedocs/conf.py
```diff
@@ -18,12 +18,12 @@
 # -- Project information -----------------------------------------------------
 
 project = 'PowerSystemDataModel'
-copyright = u'2020. TU Dortmund University, Institute of Energy Systems, Energy Efficiency and Energy Economics, Research group Distribution grid planning and operation '
+copyright = u'2023. TU Dortmund University, Institute of Energy Systems, Energy Efficiency and Energy Economics, Research group Distribution grid planning and operation '
 author = 'Johannes Hiry, Debopama Sen Sarma, Chris Kittl'
```
sebastian-peter, line 22: This could be set to `ie³` since we do not want to adapt this String all the time

docs/readthedocs/io/basiciousage.rst
```diff
@@ -10,6 +10,7 @@ Therefore, the I/O-package is structured as highly modular.
 
    influxdb
    csvfiles
+   sql
```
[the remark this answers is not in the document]
sebastian-peter: Ok I like this, maybe you could also move the validation utils page over to the io section?

sebastian-peter (changes requested):

docs/readthedocs/conf.py
staudtMarius: We don't need this. I will remove it.

docs/readthedocs/conf.py
```diff
@@ -18,12 +18,12 @@
 # -- Project information -----------------------------------------------------
 
 project = 'PowerSystemDataModel'
-copyright = u'2020. TU Dortmund University, Institute of Energy Systems, Energy Efficiency and Energy Economics, Research group Distribution grid planning and operation '
+copyright = 'ie3'
```
sebastian-peter, line 21: Please change the `auther`-field, not the `copyright`

docs/readthedocs/conf.py
sebastian-peter: @staudtMarius that's not it, yet. `version` and `release` refer to the same version number, but `version` is shorter (`3.0`).

sebastian-peter (changes requested):
Think we're getting closer!

sebastian-peter (approved):
Ok, we got it done! Thanks!

PR #714, opened 2022-12-21
Deleted parameter decimalPlaces

### LINKED ISSUE #710: Deleting decimalPlaces
The parameter decimalPlaces must be deleted

vickybung1:
Resolves #710

### DIFF
--- a/docs/readthedocs/io/csvfiles.rst
+++ b/docs/readthedocs/io/csvfiles.rst
@@ -15,13 +15,13 @@ Currently we offer two different, pre-defined naming strategies, which you might
    An extended version of the EntityPersistenceNamingStrategy. Additionally, the `Default directory hierarchy`_ is taken
    into account. Please note, that this directory hierarchy is only meant to be used in conjunction with input models.
 
-However, you can control the behaviour of serialization and de-serialization of models by injecting the desired naming
+However, you can control the behaviour of serialization and serialization of models by injecting the desired naming
 strategy you like into :code:`CsvDataSource` and :code:`CsvFileSink`.
 
 Default naming strategy
 =======================
 There is a default mapping from model class to naming of entities in the case you would like to use csv files for
-(de-)serialization of models.
+serialization of models.
 You may extend / alter the naming with pre- or suffix by calling :code:`new EntityPersistenceNamingStrategy("prefix","suffix")`.
 
 Input
@@ -194,7 +194,7 @@ However, note that this hierarchy is only meant to be used in conjunction with i
 The class :code:`DefaultInputHierarchy` offers some helpful methods to validate and create a default input file
 hierarchy.
 
-De-Serialization (loading models)
+Serialization (loading models)
 =================================
 Having an instance of :ref:`Grid Container<grid_container_model>` is most of the time the target whenever you load your
 grid. It consists of the three main blocks:
--- a/docs/readthedocs/io/csvfiles.rst
+++ b/docs/readthedocs/io/csvfiles.rst
@@ -15,13 +15,13 @@ Currently we offer two different, pre-defined naming strategies, which you might
    An extended version of the EntityPersistenceNamingStrategy. Additionally, the `Default directory hierarchy`_ is taken
    into account. Please note, that this directory hierarchy is only meant to be used in conjunction with input models.
 
-However, you can control the behaviour of serialization and serialization of models by injecting the desired naming
+However, you can control the behaviour of serialization and de-serialization of models by injecting the desired naming
 strategy you like into :code:`CsvDataSource` and :code:`CsvFileSink`.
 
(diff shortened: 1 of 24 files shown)


src/main/java/edu/ie3/datamodel/models/input/system/characteristic/CharacteristicInput.java
```diff
@@ -121,24 +113,21 @@ public SortedSet<CharacteristicPoint<A, O>> getPoints() {
   public String deSerialize() {
```
sebastian-peter, line 113: Ok maybe we can work on the naming here?

"In computing, serialization (or serialisation) is the process of translating a data structure or object state into a format that can be stored (e.g. files in secondary storage devices, data buffers in primary storage devices) or transmitted (e.g. data streams over computer networks) and reconstructed later (possibly in a different computer environment)."

"The opposite operation, extracting a data structure from a series of bytes, is deserialization, (also called unserialization or unmarshalling)."

(both from Wikipidia)

If we consider our Java objects here as the objects, and the output as a String as the "data [...] format that can be stored [...] and reconstructed later", this method should be called `serialize()`. 
But maybe we can also find a name that brings less heavy connotations with it, e.g. `export`...

@t-ober @danielfeismann

sebastian-peter (commented):
From my side, this looks all good. Don't know if you discussed any other steps last week, @danielfeismann ?

The minor point I made here could be solved in a different PR.

src/main/java/edu/ie3/datamodel/models/input/system/characteristic/CharacteristicInput.java
t-ober: I would be fine with just going for `serialize()` as this is actually pretty fitting to what we do here.

docs/readthedocs/io/csvfiles.rst
```diff
@@ -15,13 +15,13 @@ Currently we offer two different, pre-defined naming strategies, which you might
    An extended version of the EntityPersistenceNamingStrategy. Additionally, the `Default directory hierarchy`_ is taken
    into account. Please note, that this directory hierarchy is only meant to be used in conjunction with input models.
 
-However, you can control the behaviour of serialization and de-serialization of models by injecting the desired naming
+However, you can control the behaviour of serialization and serialization of models by injecting the desired naming
```
sebastian-peter, line 18: This made more sense before 😄

docs/readthedocs/io/csvfiles.rst
```diff
@@ -15,13 +15,13 @@ Currently we offer two different, pre-defined naming strategies, which you might
...
 strategy you like into :code:`CsvDataSource` and :code:`CsvFileSink`.
 
 Default naming strategy
 =======================
 There is a default mapping from model class to naming of entities in the case you would like to use csv files for
-(de-)serialization of models.
+serialization of models.
```
sebastian-peter, line 24: This should also stay as it was before. Naming strategies are used with both.

docs/readthedocs/io/csvfiles.rst
```diff
@@ -194,7 +194,7 @@ However, note that this hierarchy is only meant to be used in conjunction with i
 The class :code:`DefaultInputHierarchy` offers some helpful methods to validate and create a default input file
 hierarchy.
 
-De-Serialization (loading models)
+Serialization (loading models)
```
sebastian-peter, line 197: Was correct before

docs/uml/main/input/CharacteristicDatamodelConcept.puml
```diff
@@ -33,7 +33,7 @@ package models {
                     + {static} REQUIRED_FORMAT: String
                     - x: ComparableQuantity<A>
                     - y: ComparableQuantity<O>
-                    + deSerialize(int): String
+                    + serialize(int): String
```
sebastian-peter, line 36: You just removed the Integer parameter with this PR :)

sebastian-peter (changes requested):
Thanks for fixing serialization-related documentation and naming in so many places! I think this is a good contribution towards making the PSDM more understandable.

Please check my remarks below. Also, please enhance the changelog message with something indicating that the naming of serialization methods has changed as well.

docs/uml/main/input/CharacteristicDatamodelConcept.puml
vickybung1: I am not quite sure what's to be done here

sebastian-peter: @vickybung1 Please remove the `int` part so that it looks like the method signature in line 46

sebastian-peter (approved):
Looks good, thanks!

PR #770, opened 2023-03-06
Improving validation utils

### LINKED ISSUE #758: Improving ValidationUtils
There are two main points in class ``ValidationUtils`` :

- [x] non critical validations (e.g. geometry) should no longer throw exceptions instead their validation results should be logged.
- [x] all thrown exceptions should be collected and thrown at the end of the validation process using the ``Try`` class introduced in #712 

TO
- [x] Check uniqueness of identifiers 
- [x] Implement a connectivity check, as done in SIMONA in `GridModel.validateConnectivity` as part of the container validation utils

staudtMarius:
Resolves #758


staudtMarius:
!test

staudtMarius:
!test

src/main/java/edu/ie3/datamodel/utils/validation/ConnectorValidationUtils.java
```diff
@@ -88,15 +131,31 @@ private static void checkLine(LineInput line) {
...
+  protected static List<Try<Void, InvalidEntityException>> checkLineType(LineTypeInput lineType) {
+    try {
+      checkNonNull(lineType, "a line type");
+    } catch (InvalidEntityException e) {
+      return List.of(
+          new Failure<>(
+              new InvalidEntityException(
+                  "Validation not possible because received object {" + lineType + "} was null",
+                  e)));
+    }
```
sebastian-peter, line 146: I think this can be solved more elegantly with the existing `Try.ofVoid()`, right?
Also please check all other changed code for this opportunity to simplify the code (I've found a couple more)

src/main/java/edu/ie3/datamodel/utils/validation/ConnectorValidationUtils.java
```diff
@@ -200,42 +309,122 @@ private static void checkTransformer3W(Transformer3WInput transformer3W) {
...
    */
-  private static void checkSwitch(SwitchInput switchInput) {
-    if (!switchInput.getNodeA().getVoltLvl().equals(switchInput.getNodeB().getVoltLvl()))
-      throw new InvalidEntityException("Switch connects two different voltage levels", switchInput);
+  private static Try<Void, InvalidEntityException> checkSwitch(SwitchInput switchInput) {
+    if (!switchInput.getNodeA().getVoltLvl().equals(switchInput.getNodeB().getVoltLvl())) {
+      return Failure.ofVoid(
+          new InvalidEntityException("Switch connects two different voltage levels", switchInput));
+    } else {
+      return Success.empty();
```
sebastian-peter, line 372: Maybe for these kind of pieces of code (there are many), a 
```java
Try.ofVoid(boolean success, VoidSupplier<E> supplier, Class<E> clazz)
```
would be advantageous.

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -125,27 +183,79 @@ else if (ThermalUnitInput.class.isAssignableFrom(assetInput.getClass()))
...
+      Set<? extends AssetInput> inputs) {
+    List<String> ids = new ArrayList<>();
+    List<Try<Void, UnsafeEntityException>> exceptions = new ArrayList<>();
+
+    inputs.forEach(
+        input -> {
+          String id = input.getId();
+          if (!ids.contains(id)) {
+            ids.add(id);
+            exceptions.add(Success.empty());
```
sebastian-peter, line 249: Since we're not really interested in the Void-successes produced during validation, maybe we can skip adding these here?

src/main/java/edu/ie3/datamodel/utils/validation/GridContainerValidationUtils.java
```diff
@@ -86,26 +133,38 @@ protected static void checkRawGridElements(RawGridElements rawGridElements) {
...
         .forEach(
             line -> {
-              checkNodeAvailability(line, nodes);
-              ConnectorValidationUtils.check(line);
+              try {
+                checkNodeAvailability(line, nodes);
+              } catch (InvalidGridException e) {
+                exceptions.add(new Failure<>(e));
+              }
+              exceptions.addAll(ConnectorValidationUtils.check(line));
```
sebastian-peter, line 141: These could also be done with `Try.ofVoid()`?

src/main/java/edu/ie3/datamodel/utils/Try.java
sebastian-peter, line 1: One point about `Try.Success#empty` that I just noticed: You could create the `new Success<>(null)` once and save it in a static field, so that this object only has to be created once.

src/main/java/edu/ie3/datamodel/utils/validation/ConnectorValidationUtils.java
```diff
@@ -38,21 +50,41 @@ private ConnectorValidationUtils() {
...
-    connectsDifferentNodes(connector);
+  protected static List<Try<Void, InvalidEntityException>> check(ConnectorInput connector) {
+    try {
+      checkNonNull(connector, "a connector");
+    } catch (InvalidEntityException e) {
+      return List.of(
+          new Failure<>(
+              new InvalidEntityException(
+                  "Validation not possible because received object {" + connector + "} was null",
+                  e)));
```
sebastian-peter, line 64: This exception is built a lot of times, and SQ is already complaining about the duplicate String literals. Maybe creating a small static method for this would make sense, maybe in `ValidationUtils`?

sebastian-peter (changes requested):
Hopefully, with the suggestions I have here the additional lines of code can shrink done a bit again :)

Furthermore, I think we should elevate the log messages to at least info, better yet warn. Debug is not enough.

Please also check out the [SonarQube code smells for this PR](https://simona.ie3.e-technik.tu-dortmund.de/sonar/project/issues?pullRequest=770&resolved=false&types=CODE_SMELL&inNewCodePeriod=true&id=edu.ie3%3APowerSystemDataModel).

staudtMarius:
!test

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -54,18 +59,39 @@ protected static NotImplementedException checkNotImplementedException(Object obj
...
-    checkNonNull(obj, "an object");
-    if (AssetInput.class.isAssignableFrom(obj.getClass())) checkAsset((AssetInput) obj);
-    else if (GridContainer.class.isAssignableFrom(obj.getClass()))
-      GridContainerValidationUtils.check((GridContainer) obj);
-    else if (GraphicInput.class.isAssignableFrom(obj.getClass()))
-      GraphicValidationUtils.check((GraphicInput) obj);
-    else if (AssetTypeInput.class.isAssignableFrom(obj.getClass()))
-      checkAssetType((AssetTypeInput) obj);
-    else throw checkNotImplementedException(obj);
+  public static Try<Void, ? extends ValidationException> check(Object obj) {
```
sebastian-peter, line 65: It'd be great if, like with inputs, we don't force the `Try` class on users of this validation functionality. So I'd rather prefer a method signature like this:
```java
public static void check(Object obj) throws ValidationException
```
This is also important for all the tests using `noExceptionThrown()` to work.

On that note, `ValidationException` should really _not_ be a `RuntimeException`, so that `throws ...` has to be declared wherever we actually throw.

src/test/groovy/edu/ie3/datamodel/utils/validation/NodeValidationUtilsTest.groovy
```diff
@@ -27,15 +27,15 @@ class NodeValidationUtilsTest extends Specification {
     def node = GridTestData.nodeA
 
     when:
-    NodeValidationUtils.check(node)
+    List<Try<Void, ? extends ValidationException>> tries = NodeValidationUtils.check(node)
 
     then:
-    noExceptionThrown()
+    tries.forEach { it.success }
```
sebastian-peter, line 33: Use `every` here, otherwise it won't fail. I also found a bunch of tests from the previous PR using `forEach`, please fix them as well.

sebastian-peter (changes requested):
Two small points remain...

src/test/groovy/edu/ie3/datamodel/utils/validation/ValidAssetInput.groovy
```diff
@@ -0,0 +1,27 @@
...
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+ */
+package edu.ie3.datamodel.utils.validation
+
+import edu.ie3.datamodel.models.input.AssetInput
+
+import java.time.ZonedDateTime
+
+class ValidAssetInput extends AssetInput {
```
sebastian-peter, line 12: I think the classes `ValidAssetInput` and `InvalidAssetInput` could be unified into one, maybe `TestAssetInput` or `DummyAssetInput`
In this instance, I think it would make the tests that they are involved in even clearer.

sebastian-peter (changes requested):
There's also impractical usages of `forEach` in `CsvRawGridSourceTest`, could you fix them as well?

src/main/java/edu/ie3/datamodel/models/input/container/InputContainer.java
```diff
@@ -34,6 +35,6 @@ protected InputContainerCopyBuilder() {}
     protected abstract InputContainerCopyBuilder<T, E> childInstance();
```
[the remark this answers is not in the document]
staudtMarius: Maybe we can make this method not abstract? This way we could reduce some duplicate Code?

staudtMarius:
!test

staudtMarius:
!test

staudtMarius:
!test

sebastian-peter (dismissed):
Ok, we got it know imho. I only adapted some minor stuff. Thanks for doing the large part of the job, @staudtMarius!

sebastian-peter (approved):

PR #771, opened 2023-03-07
Changed switch result documentation according to the implementation

### LINKED ISSUE #757: Fix switch result documentation
[Switch result documentation](https://powersystemdatamodel.readthedocs.io/en/latest/models/result/grid/switch.html) shows different attributes to what the implementation (`edu.ie3.datamodel.models.result.connector.SwitchResult`) actually offers.

Adapt the documentation so that it fits the implementation.

(If I had to make a guess, the error probably stems from the erroneous assumption that SwitchResults are ConnectorResults)

vickybung1:
Resolves #757


danielfeismann (approved):
Looks good, thanks for updating the documentation!

PR #792, opened 2023-04-11
Deleted initFiles. Deleted all initFiles Tests. Set parameter append to false by default

### LINKED ISSUE #791: Delete option to init files without providing any elements
Instead of including an option to overwrite files (#763) initFiles should be deleted and append should by set to false by default.

vickybung1:
Resolves #791


src/test/groovy/edu/ie3/datamodel/io/sink/CsvFileSinkTest.groovy
```diff
@@ -140,91 +113,6 @@ class CsvFileSinkTest extends Specification implements TimeSeriesTestData {
...
-    new File(testBaseFolderPath + File.separator + "operator_input.csv").exists()
-    new File(testBaseFolderPath + File.separator + "node_graphic_input.csv").exists()
-    new File(testBaseFolderPath + File.separator + "thermal_bus_input.csv").exists()
-    new File(testBaseFolderPath + File.separator + "thermal_house_input.csv").exists()
-    new File(testBaseFolderPath + File.separator + "load_input.csv").exists()
-    new File(testBaseFolderPath + File.separator + "em_input.csv").exists()
-
-    !new File(testBaseFolderPath + File.separator + "ev_res.csv").exists()
-  }
-
```
t-ober, line 227: We can and should keep the test. We can just remove the `initFiles` parameter and adjust the test name description to `A valid CsvFileSink without 'initFiles' should persist provided elements correctly`

t-ober (changes requested):
Looks good to me. I would just revive the mentioned test and we need address the merge conflict

t-ober (dismissed):

t-ober (approved):

PR #794, opened 2023-04-17
Allow hierarchic grid structure for JointGridContainer

### LINKED ISSUE #768: JointGridContainerSource should be able to read hierarchic grids
Currently we assume a flat directory hierarchy.

jo-bao:
Resolves #768


jo-bao:
!test

jo-bao:
!test

src/main/java/edu/ie3/datamodel/models/input/container/InputContainer.java
```diff
@@ -15,6 +15,11 @@
   /** @return unmodifiable List of all entities */
   List<T> allEntitiesAsList();
 
+  /** Returns number of holded elements in the container. */
+  default int numberOfElements() {
+    return allEntitiesAsList().size();
+  }
+
```
t-ober, line 22: This method is never used. Maybe we don't need it. 

Also "holds"

jo-bao: Oh yes, in this implementation it isn't necessary. I'll remove it.
As a remark: In failing groovy tests all list elements will be displayed. Then the error message is very big and messy. This method was a way to avoid that.

t-ober (approved):

PR #796, opened 2023-04-18
Removing support for old csv format

### LINKED ISSUE #795: Remove support for old CSV fromat
As found out in issue #540 there are some tests that fail after removing support for the old CSV format. That is way the support was not removed.

staudtMarius:
Resolves #795


docs/readthedocs/gettingstarted.md
sebastian-peter, line 1: Nice!

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -90,6 +81,11 @@ public Stream<Map<String, String>> getSourceData(Class<? extends Entity> entityC
 
   // -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
 
+  /** Returns the set {@link FileNamingStrategy}. */
+  public FileNamingStrategy getNamingStrategy() {
+    return fileNamingStrategy;
+  }
+
```
sebastian-peter, line 88: Is this on purpose?

sebastian-peter (commented):
Just a question from my side

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
staudtMarius: Yes. Befor we had a public method between protected methods and I thought it is better to have all public methods next to each other.

sebastian-peter: Ah, I overlooked this. Good.

sebastian-peter (approved):

PR #798, opened 2023-04-24
Make ``handleEntity`` throw exceptions instead of returning empty optional

### LINKED ISSUE #682: Make the method `handleEntity` throw an Exception if it catches one
Especially since `Processor.processMethodResults` tends to throw exceptions when handling of specific fields isn't dealt with we want the program to crash instead of having to search for why we get an optional

staudtMarius:
Resolves #682


sebastian-peter (changes requested):
Thanks for the contribution. Here's some things I noticed.

sebastian-peter (approved):
I only had some minor adaptations from my side, looks very good overall!

PR #799, opened 2023-04-24
Included option to parse a DateTimeFormatter to the processors

### LINKED ISSUE #766: Use the default time pattern for persisting time series keys instead of the `toString` method
Currently the to string persistence outputs time strings with the [UTC] postfix which we can't easily parse. Use the default time pattern instead.

vickybung1:
Resolves #766


jo-bao:
!test

sebastian-peter:
Closing this in favor of #767, which uses a fixed pattern and causes less "parameter pass-through"

PR #811, opened 2023-05-16
Enforcing correct assignment of transformer nodes to the voltage side

### LINKED ISSUE #803: Transformer two-winding hv and lv node validation
Our documentation states that `nodeA` of a transformer should be the hv-side and `nodeB` the lv-side. 
To me it seems like we do not enforce that in any way, which we should.
It would be nice to check that during object construction and also add this to the `ValidationUtils`.

vickybung1:
Resolves #803

### DIFF
--- a/src/test/groovy/edu/ie3/datamodel/utils/validation/ConnectorValidationUtilsTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/utils/validation/ConnectorValidationUtilsTest.groovy
@@ -190,7 +190,7 @@ class ConnectorValidationUtilsTest extends Specification {
     GridTestData.transformerAtoBtoC.copy().tapPos(100).build()                                                                  		|| 1            || new InvalidEntityException("Tap position of Transformer3WInput is outside of bounds", invalidTransformer3W)
     GridTestData.transformerAtoBtoC.copy().nodeA(GridTestData.nodeA.copy().voltLvl(GermanVoltageLevelUtils.HV).build()).build() 		|| 2            || new InvalidEntityException("Transformer connects nodes of the same voltage level", invalidTransformer3W)
     GridTestData.transformerAtoBtoC.copy().nodeA(GridTestData.nodeA.copy().subnet(2).build()).build()                           		|| 1            || new InvalidEntityException("Transformer connects nodes in the same subnet", invalidTransformer3W)
-    GridTestData.transformerAtoBtoC.copy().nodeA(GridTestData.nodeA.copy().voltLvl(GermanVoltageLevelUtils.MV_30KV).build()).build() 	|| 1            || new InvalidEntityException("Rated voltages of Transformer3WInput do not equal voltage levels at the nodes", invalidTransformer3W)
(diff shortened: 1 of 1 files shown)


src/main/java/edu/ie3/datamodel/io/factory/input/Transformer2WInputFactory.java
```diff
@@ -37,6 +37,13 @@ protected Transformer2WInput buildModel(
       NodeInput nodeB,
       OperatorInput operator,
       OperationTime operationTime) {
+
+    if (nodeB.getVoltLvl().getNominalVoltage().compareTo(nodeA.getVoltLvl().getNominalVoltage())
+        > 0) {
+      throw new IllegalArgumentException(
+          "nodeA must be on the upper voltage side of the transformer");
+    }
```
t-ober, line 45: I would move that into the actual constructor of the class `Transformer2WInput`. Since there are different constructors make sure that the method gets called in each of them.

src/main/java/edu/ie3/datamodel/utils/validation/GridContainerValidationUtils.java
```diff
@@ -343,4 +359,19 @@ private static InvalidGridException getMissingNodeException(AssetInput input) {
...
+   * @param input Input model
+   * @return Exception for the wrong voltage side
+   */
+  private static InvalidGridException getWrongVoltageSideException(AssetInput input) {
+    return new InvalidGridException(
+        input.getClass().getSimpleName()
+            + "The nodes of transformer"
+            + input
+            + " are located at the wrong voltage side");
+  }
```
t-ober, line 376: Since we only use it once we for now I think we don't need a more generic method here. Also you could add that we expect the node a to have the higher voltage level into the error message.

t-ober (changes requested):
Thanks, I have some minor suggestions!

t-ober (changes requested):
Thanks, I have some minor suggestions!

src/test/groovy/edu/ie3/datamodel/io/factory/input/Transformer2WInputFactoryTest.groovy
```diff
@@ -63,4 +64,29 @@ class Transformer2WInputFactoryTest extends Specification implements FactoryTest
...
+      "operatesfrom"   : "2019-01-01T00:00:00+01:00[Europe/Berlin]",
+      "operatesuntil"  : "",
+      "id"             : "TestID",
+      "paralleldevices": "2",
+      "tappos"         : "3",
+      "autotap"        : "true"
+    ]
+    def inputClass = Transformer2WInput
+    def operatorInput = Mock(OperatorInput)
+    def nodeInputA = GridTestData.nodeB
```
danielfeismann, line 81: Do we want to accept transformers with the same voltage on both sides? If yes this test need to be adapted that we then don't expect an exception. If not the method itself should check for greaterThanOrEqual. Same for three- winding transformer.

danielfeismann (changes requested):
Looks good, just one question.

src/test/groovy/edu/ie3/datamodel/io/factory/input/Transformer2WInputFactoryTest.groovy
t-ober: As per our discussion we allow the equal case.

danielfeismann (approved):
LGTM

PR #818, opened 2023-05-26
Fix wrong hashcode for ConnectorResult

### LINKED ISSUE #817: The Hashcode of results should not be the same if they belong to different input models
Currently when having e.g. two `Transformer2WResult` from different transformers their hash is the same so one of them gets discarded when you put them both into the same set. This should not happen.

t-ober:
Resolves #817


sebastian-peter (approved):
Looks good, important fix, thanks

PR #926, opened 2023-11-08
Improving column name validation

### LINKED ISSUE #670: Enhance error handling of coordinate source with invalid column names
When we have a coordinate source with invalid column names we don't get a useful error message but a `NullPointerException`. We should be able to give more sensible feedback by first checking if these columns exist or throwing a more useful exception.

```
java.lang.NullPointerException: Cannot invoke "String.concat(String)" because the return value of "java.util.Map.get(Object)" is null
	at edu.ie3.datamodel.io.source.csv.CsvIdCoordinateSource.lambda$buildStreamWithFieldsToAttributesMap$2(CsvIdCoordinateSource.java:123)
	at edu.ie3.datamodel.utils.validation.ValidationUtils.lambda$distinctByKey$5(ValidationUtils.java:245)
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:178)
	at java.base/java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1707)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:960)
	at java.base/java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:934)
	at java.base/java.util.stream.AbstractTask.compute(AbstractTask.java:327)
	at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:754)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:686)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:927)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at edu.ie3.datamodel.io.source.csv.CsvDataSource.distinctRowsWithLog(CsvDataSource.java:454)
	at edu.ie3.datamodel.io.source.csv.CsvIdCoordinateSource.buildStreamWithFieldsToAttributesMap(CsvIdCoordinateSource.java:124)
	at edu.ie3.datamodel.io.source.csv.CsvIdCoordinateSource.setupIdToCoordinateMap(CsvIdCoordinateSource.java:52)
	at edu.ie3.datamodel.io.source.csv.CsvIdCoordinateSource.<init>(CsvIdCoordinateSource.java:42)
	at edu.ie3.simona.service.weather.WeatherSource$.$anonfun$checkCoordinateSource$3(WeatherSource.scala:460)
	at edu.ie3.simona.service.weather.WeatherSourceWrapper$.apply(WeatherSourceWrapper.scala:321)
	at edu.ie3.simona.service.weather.WeatherSource$.$anonfun$checkConfig$6(WeatherSource.scala:390)
	at edu.ie3.simona.service.weather.WeatherSource$.apply(WeatherSource.scala:295)
	at edu.ie3.simona.service.weather.WeatherService.init(WeatherService.scala:136)
	at edu.ie3.simona.service.SimonaService$$anonfun$uninitialized$1.applyOrElse(SimonaService.scala:64)
	at akka.actor.Actor.aroundReceive(Actor.scala:537)
	at akka.actor.Actor.aroundReceive$(Actor.scala:535)
	at edu.ie3.simona.service.SimonaService.aroundReceive(SimonaService.scala:36)
	at akka.actor.ActorCell.receiveMessage(ActorCell.scala:579)
	at akka.actor.ActorCell.invoke(ActorCell.scala:547)
	at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:270)
	at akka.dispatch.Mailbox.run(Mailbox.scala:231)
	at akka.dispatch.Mailbox.exec(Mailbox.scala:243)
```

### LINKED ISSUE #839: Allow additional unused columns in csv files
As defined in the `Factory.validateParameters()` method we filter on a set of attributes that exactly match the expected parameters, also throwing an exception if there are additional fields. Since we can just ignore them I do not think we need to throw an exception, but rather just log that they will not be used. 

This maybe up for discussion, but I think the added flexibility might very well be worth it.
@sebastian-peter : Thoughts?

### LINKED ISSUE #849: Column name validation should run only once: early during initialization
This came up originally in #807, which is only a quick fix for part 1 of the problem:

Ok I just realized we already have a validation in `Factory#validateParameters`, which is called by `Factory#get(D)`, which is called everytime that any entity is parsed from some source. The problem with IdCoordinateSources is that these validations are called _after_ some other things, like duplicate removal, happened, that already access the data and fail if some column is missing.

Looking at `Factory.get()` as a whole, I'd describe two main issues:
1. validation is sometimes, like in the case of `IdCoordinateSource`, called too late
2. validation is executed for _every single_ data point parsed from `Factory.get()`, which is pretty costly

So I think there's a way to fix both at the same time: Validate column names once during initialization (and early enough).

_Originally posted by @sebastian-peter in https://github.com/ie3-institute/PowerSystemDataModel/issues/807#issuecomment-1651623738_

staudtMarius:
Resolves #670
Resolves #839
Resolves #849


src/main/java/edu/ie3/datamodel/io/source/DataSource.java
```diff
@@ -13,5 +13,6 @@
 public interface DataSource {
 
   /** Creates a stream of maps that represent the rows in the database */
-  Stream<Map<String, String>> getSourceData(Class<? extends UniqueEntity> entityClass);
+  Stream<Map<String, String>> getSourceData(
+      Class<? extends UniqueEntity> entityClass, SourceValidator validateSource);
```
sebastian-peter, line 17: I'm thinking out loud here, but maybe it'd be better to have validation happen separately from creating the field mappings? It doesn't really fit here, logically.


I know, the price to pay here would be to read out the source file twice (once only for the headline, second time for the content), but honestly I think that's still a cheap price if we can keep the validation less entangled from regular factory functions here.

sebastian-peter (changes requested):
The review took a bit longer than expected, because a lot of thoughts came up on my side. Let me know what you think, and whether I can assist.

src/main/java/edu/ie3/datamodel/io/factory/Factory.java
```diff
@@ -108,76 +104,87 @@ private void isSupportedClass(Class<?> desiredClass) {
...
+   * @param foundFields that were found
+   * @param entityClass of the build data
    */
-  protected int validateParameters(D data, Set<String>... fieldSets) {
-    Map<String, String> fieldsToValues = data.getFieldsToValues();
-
-    // get all sets that match the fields to attributes
+  public void validate(Set<String> foundFields, Class<?> entityClass) {
+    List<Set<String>> fieldSets =
+        getFields(entityClass).stream().map(Factory::toSnakeCase).toList();
```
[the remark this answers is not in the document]
staudtMarius: If we change the constant fields, we would have either to convert all fields found in a source to snake case or enforce all source to be in snake case. With the current implementation either snake or camel cases are supported.

src/main/java/edu/ie3/datamodel/io/factory/Factory.java
```diff
@@ -48,15 +49,10 @@ public List<Class<? extends C>> getSupportedClasses() {
...
-    // magic: case-insensitive get/set calls on set strings
-    final List<Set<String>> allFields = getFields(data);
-
     try {
-      validateParameters(data, allFields.toArray((IntFunction<Set<String>[]>) Set[]::new));
-
       // build the model
       return Success.of(buildModel(data));
-    } catch (FactoryException e) {
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
staudtMarius: Inside the `buildModel` method other exceptions like `IllegalArgumentException` can be thrown. These exceptions need to be catched, because otherwise no `Try` is returned. A possible solution would be to catch all types of exceptions that coulb be thrown.

staudtMarius:
!test

src/main/java/edu/ie3/datamodel/io/connectors/CsvFileConnector.java
```diff
@@ -51,6 +57,59 @@ public CsvFileConnector(Path baseDirectoryName, FileNamingStrategy fileNamingStr
     this.fileNamingStrategy = fileNamingStrategy;
   }
 
+  @SuppressWarnings("unchecked")
+  public void validate(Map<Class<?>, SourceValidator<?>> pairs, String csvSep) {
+    List<FactoryException> exceptions = new ArrayList<>();
+
```
sebastian-peter, line 63: If we want to match the entity class with the validator type class, we should not expect a map as an argument, but just a single class/validator combination like within other connectors.

Secondly, how about making csvSep part of the `CsvFileConnector` class (passed via constructor). This way, all `validate` methods have the same method signature and can be required by `DataSource`.
Or, alternatively to this second point:
I think semantically it'd make more sense to have the `validate` method inside the `DataSource`s (like `CsvDataSource`) instead of the `DataConnector`s (like `CsvFileConnector`). So there would be a `validate` method required by `DataSource`. Would this be possible?

Finally we can then call the validate method from `*Source` rather than `Csv*Source` (for most entities). This will also come in handy when SQL sources are introduced.

src/main/java/edu/ie3/datamodel/io/connectors/InfluxDbConnector.java
```diff
@@ -107,6 +110,26 @@ public InfluxDbConnector(String url, String databaseName) {
     this(url, databaseName, NO_SCENARIO, true, InfluxDB.LogLevel.NONE, BatchOptions.DEFAULTS);
   }
 
+  /**
+   * This method should be used to validate a given {@link InfluxDB}.
+   *
+   * @param entityClass class of the entity
+   * @param validator for validation
+   */
+  public final void validateDb(Class<?> entityClass, SourceValidator validator) {
```
sebastian-peter, line 119: Same as with the other connectors

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -179,6 +179,9 @@ private Stream<Map<String, String>> buildStreamWithFieldsToAttributesMap(
     try (BufferedReader reader = bufferedReader) {
       final String[] headline = dataSource.parseCsvRow(reader.readLine(), dataSource.csvSep);
 
+      // validating read file
+      weatherFactory.validate(Set.of(headline), WeatherValue.class);
```
sebastian-peter, line 183: Why not use the method in `CsvFileConnector` (or `CsvDataSource`)?

src/main/java/edu/ie3/datamodel/io/source/csv/CsvIdCoordinateSource.java
```diff
@@ -162,6 +162,9 @@ protected Stream<Map<String, String>> buildStreamWithFieldsToAttributesMap() {
     try (BufferedReader reader = dataSource.connector.initIdCoordinateReader()) {
       final String[] headline = dataSource.parseCsvRow(reader.readLine(), dataSource.csvSep);
 
+      // validating read file
+      factory.validate(Set.of(headline), Pair.class);
```
sebastian-peter, line 166: Same here

sebastian-peter (changes requested):

src/main/java/edu/ie3/datamodel/io/source/csv/CsvIdCoordinateSource.java
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/io/connectors/CsvFileConnector.java
staudtMarius: It is not possible to put the validation method inside the `DataSource`, because we need the information from the `Connectors`.
Also if we have the validation inside the `Connector`s we don't need to implement validation methods for each type of `Source` (like `DataSource`, `IdCoordinateSource`, ....).

sebastian-peter:
Reminder about open question: Validation in `TimeSeriesMetaInformationSource`?

sebastian-peter (approved):

PR #931, opened 2023-11-20
Abstracting some method in `ValidationUtils`.

### LINKED ISSUE #852: Abstract `checkIds` and `checkForDuplicateUuids` to common method
In `ValidationUtils`, these methods work in a very similar fashion and can probably be abstracted and unified into one method.

To achieve this, maybe defining a function parameter like this could work:

```java
public interface FieldSupplier<A, B extends UniqueEntity> {
  A getField(B entity);
}
```

staudtMarius:
Resolves #852

### DIFF
--- a/src/main/java/edu/ie3/datamodel/exceptions/DuplicateEntitiesException.java
+++ b/src/main/java/edu/ie3/datamodel/exceptions/DuplicateEntitiesException.java
@@ -1,0 +1,27 @@
+/*
+ * © 2023. TU Dortmund University,
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.exceptions;
+
+import edu.ie3.datamodel.models.UniqueEntity;
+import edu.ie3.datamodel.utils.ExceptionUtils;
+import java.util.Collection;
+
+public class DuplicateEntitiesException extends ValidationException {
+
+  protected DuplicateEntitiesException(String s) {
+    super(s);
+  }
+
+  protected DuplicateEntitiesException(String s, String entities) {
+    super(s + entities);
+  }
+
+  public DuplicateEntitiesException(String fieldName, Collection<? extends UniqueEntity> entities) {
+    this(
+        "The following entities have duplicate '" + fieldName + "':",
+        ExceptionUtils.combine(entities));
+  }
+}
--- a/src/main/java/edu/ie3/datamodel/exceptions/DuplicateEntitiesException.java
+++ b/src/main/java/edu/ie3/datamodel/exceptions/DuplicateEntitiesException.java
@@ -21,7 +21,7 @@ public class DuplicateEntitiesException extends ValidationException {
 
   public DuplicateEntitiesException(String fieldName, Collection<? extends UniqueEntity> entities) {
     this(
-        "The following entities have duplicate '" + fieldName + "':",
+        "The following entities have duplicate '" + fieldName + "': ",
         ExceptionUtils.combine(entities));
   }
 }
--- a/src/main/java/edu/ie3/datamodel/utils/ExceptionUtils.java
+++ b/src/main/java/edu/ie3/datamodel/utils/ExceptionUtils.java
@@ -5,7 +5,10 @@
 */
 package edu.ie3.datamodel.utils;
 
+import edu.ie3.datamodel.models.UniqueEntity;
+import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class ExceptionUtils {
   private ExceptionUtils() {
@@ -23,5 +26,15 @@ public class ExceptionUtils {
         .map(Throwable::getMessage)
         .reduce("", (a, b) -> a + "\n " + b)
         .replaceFirst("\n ", "");
+  }
+
+  /**
+   * Combines multiple {@link UniqueEntity} into a string.
+   *
+   * @param entities to be combined
+   * @return a string
+   */
+  public static String combine(Collection<? extends UniqueEntity> entities) {
+    return entities.stream().map(UniqueEntity::toString).collect(Collectors.joining("\n - "));
   }
 }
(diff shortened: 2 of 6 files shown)


staudtMarius:
!test

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -354,38 +326,44 @@ public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor
...
-   * @param entities the entities that should be checkd for UUID uniqueness
-   * @return either a string wrapped in an optional with duplicate UUIDs or an empty optional
+   * @param entities to be checked
+   * @param supplier for the field
+   * @return a list of {@link Try}.
+   * @param <A> type of the field
+   * @param <B> type of the {@link UniqueEntity}
    */
-  protected static Optional<String> checkForDuplicateUuids(Set<UniqueEntity> entities) {
-    if (distinctUuids(entities)) {
```
sebastian-peter, line 364: The `distinctUuids` method is now also obsolete

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -354,38 +326,44 @@ public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor
...
-                  return entry.getKey()
-                      + ": "
-                      + entry.getValue()
-                      + "\n - "
-                      + duplicateEntitiesString;
-                })
-            .collect(Collectors.joining("\n\n"));
-
-    return Optional.of(duplicationsString);
+  protected static <A, B extends UniqueEntity>
```
sebastian-peter, line 337: I think the plural form `checkForDuplicates` fits better

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -354,38 +326,44 @@ public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor
...
+                  Failure.ofVoid(
+                      new DuplicateEntitiesException(
+                          key.getClass().getSimpleName(), duplicates.get(key))));
+            });
+
+    return exceptions;
+  }
+
+  /**
+   * Supplier for unique entity fields.
```
sebastian-peter, line 360: "... that returns a field value of type A given an entity of type B"

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -354,38 +326,44 @@ public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor
...
+   *
+   * @param <A> type of field
+   * @param <B> type of unique entity
+   */
+  @FunctionalInterface
+  protected interface FieldSupplier<A, B extends UniqueEntity> {
```
sebastian-peter, line 366: Following conventional (mathematical) function definitions, I'd switch A and B, since your "input" is of type B and your "output" is of type A. I'd rather have A -> B.

But maybe we can rename them further, B could become E (entity) while A could become F (field). So: E -> F

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -354,38 +326,44 @@ public static <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor
...
+        .forEach(
+            duplicate -> {
+              A key = duplicate.getKey();
+              exceptions.add(
+                  Failure.ofVoid(
+                      new DuplicateEntitiesException(
+                          key.getClass().getSimpleName(), duplicates.get(key))));
+            });
+
+    return exceptions;
```
sebastian-peter, line 356: Love everything about the way you arranged this, just the `map` operation that was used before, I think is more neat.

This way, you don't need the intermediary `exceptions` list.

I tried it here, let me know what you think:

```suggestion
    return entities.stream().collect(Collectors.groupingBy(supplier::getField)).entrySet().stream()
        .filter(e -> e.getValue().size() > 1)
        .map(
            duplicate ->
                Failure.ofVoid(
                    new DuplicateEntitiesException(
                        duplicate.getKey().getClass().getSimpleName(), duplicate.getValue())))
        .collect(Collectors.toList());
```

src/main/java/edu/ie3/datamodel/exceptions/DuplicateEntitiesException.java
```diff
@@ -0,0 +1,27 @@
...
+    super(s);
+  }
+
+  protected DuplicateEntitiesException(String s, String entities) {
+    super(s + entities);
+  }
+
+  public DuplicateEntitiesException(String fieldName, Collection<? extends UniqueEntity> entities) {
+    this(
+        "The following entities have duplicate '" + fieldName + "':",
```
sebastian-peter, line 24:
```suggestion
        "The following entities have duplicate '" + fieldName + "': ",
```

src/main/java/edu/ie3/datamodel/utils/ExceptionUtils.java
```diff
@@ -24,4 +27,14 @@ public static String getMessages(List<? extends Exception> exceptions) {
...
   }
+
+  /**
+   * Combines multiple {@link UniqueEntity} into a string.
+   *
+   * @param entities to be combined
+   * @return a string
+   */
+  public static String combine(Collection<? extends UniqueEntity> entities) {
+    return entities.stream().map(UniqueEntity::toString).collect(Collectors.joining("\n - "));
```
sebastian-peter, line 38: The first item is placed in the same line, all following items are placed in consecutive lines. Maybe you can find a better solution here.

sebastian-peter (changes requested):
Some minor remarks from my side. If you like, you can directly "commit" the suggestions I made.

PR #932, opened 2023-11-20
Compare UUIDs instead of Strings.

### LINKED ISSUE #829: Compare UUIDs instead of Strings in `EntitySource.findFirstEntityByUuid`
Once #773 is merged:

This could be a quick one. I think it's more efficient to compare UUIDs in `findFirstEntityByUuid` instead of Strings (UUIDs just consist of two longs).

_Originally mentioned by @sebastian-peter in https://github.com/ie3-institute/PowerSystemDataModel/pull/773#discussion_r1211379308_

staudtMarius:
Resolves #829

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/source/EntitySource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/EntitySource.java
@@ -71,10 +71,10 @@ public abstract class EntitySource {
    *     optional if no matching entity with the provided uuid can be found
    */
   protected <T extends UniqueEntity> Optional<T> findFirstEntityByUuid(
-      String entityUuid, Collection<T> entities) {
+      UUID entityUuid, Collection<T> entities) {
     return entities.stream()
         .parallel()
-        .filter(uniqueEntity -> uniqueEntity.getUuid().toString().equalsIgnoreCase(entityUuid))
+        .filter(uniqueEntity -> uniqueEntity.getUuid().equals(entityUuid))
         .findFirst();
   }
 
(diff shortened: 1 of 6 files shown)


staudtMarius:
!test

src/main/java/edu/ie3/datamodel/io/source/EntitySource.java
```diff
@@ -71,10 +71,10 @@ protected String safeMapGet(Map<String, String> map, String key, String mapName)
    *     optional if no matching entity with the provided uuid can be found
    */
   protected <T extends UniqueEntity> Optional<T> findFirstEntityByUuid(
-      String entityUuid, Collection<T> entities) {
+      UUID entityUuid, Collection<T> entities) {
     return entities.stream()
         .parallel()
-        .filter(uniqueEntity -> uniqueEntity.getUuid().toString().equalsIgnoreCase(entityUuid))
+        .filter(uniqueEntity -> uniqueEntity.getUuid().equals(entityUuid))
```
sebastian-peter, line 77: Okay what one should actually do (at least for frequently searched entity sets like nodes) is build a HashMap uuid -> entity, which is way faster. But that's way beyond the scope of this PR.

sebastian-peter (changes requested):
Two tiny remarks from my side.

src/main/java/edu/ie3/datamodel/io/source/EntitySource.java
staudtMarius: Should I change something in this line or are you creating a new issue for this?

sebastian-peter: Honestly I'd just leave it as it is for now, and I don't think we need an issue for this yet. If we want to optimize for speed at some point, we should measure performance (with profiling or such) anyways.

sebastian-peter (approved):
Thanks, looks good!

PR #933, opened 2023-11-20
JavaDoc cautionary hint for `safeMapGet`.

### LINKED ISSUE #828: Throw exception instead of questionable `EntitySource.safeMapGet`
Once #773 is merged:

The whole procedure within `EntitySource.safeMapGet` seems somewhat questionable to me. We rather want to throw an exception if a required value does not exist, instead of returning some error message as a String in place of the desired value. For non-required values (if applicable), one can just handle the result of `Optional.ofNullable(map.get(key))`.

_Originally mentioned by @sebastian-peter in https://github.com/ie3-institute/PowerSystemDataModel/pull/773#discussion_r1211372909_

staudtMarius:
Resolves #828

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 ### Changed
+- Replacing `EntitySource.safeMapGet` by `EntitySource.getOrThrow` [#828](https://github.com/ie3-institute/PowerSystemDataModel/issues/828)
 
 ## [4.1.0] - 2023-11-02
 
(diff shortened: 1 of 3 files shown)


sebastian-peter (changes requested):
Ok, upon inspection, I now understand the purpose of `safeMapGet`. It's only used in creating warning messages, where we do not want to fail if a value is missing (I guess this is also why you sometimes used `getOrDefault` as a replacement).

I now think it's best if we keep everything as it is, _execpt_ placing a warning within JavaDoc above safeMapGet: It should only be used for building warning messages, and unintended results can be expected if used in other contexts.

sebastian-peter (approved):
Thanks!

PR #984, opened 2024-01-08
Fixing `CsvSystemParticipantSourceTest`

### LINKED ISSUE #958: Fix `CsvSystemParticipantSourceTest`
`CsvSystemParticipantSourceTest` is broken because the tests do not fail for neither `Try.Success` or `Try.Failure`. So, if all parsing fails for whatever reason, the tests do not fail.
What could solve this is another parameter for each test `shouldSucceed`, so that we can match against expected successes and failures.

staudtMarius:
Resolves #958

### DIFF
--- a/src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
@@ -80,25 +80,58 @@ class CsvSystemParticipantSourceTest extends Specification implements CsvTestDat
     "Linked node with UUID 4ca90220-74c2-4369-9afa-a18bf068840d was not found for entity AssetInputEntityData")
   }
 
-  def "A SystemParticipantSource with csv input should return data from a valid heat pump input file as expected"() {
+  def "A SystemParticipantSource with csv input should return data from valid input file as expected"() {
     given:
     def csvSystemParticipantSource = new SystemParticipantSource(
     Mock(TypeSource),
     Mock(ThermalSource),
     Mock(RawGridSource),
     new CsvDataSource(csvSep, participantsFolderPath, fileNamingStrategy))
-    def nodeMap = map([sptd.hpInput.node])
+    def nodeMap = map([sptd.participantNode])
+    def operatorMap = map([sptd.operator])
+    def thermalBusMap = map([sptd.thermalBus])
+    def thermalStorageMap = map([sptd.thermalStorage])
 
     expect:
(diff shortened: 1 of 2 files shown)


src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
```diff
@@ -92,23 +92,39 @@ class CsvSystemParticipantSourceTest extends Specification implements CsvTestDat
...
+    def nodeMap = map([sptd.hpInput.node])
+
+    expect:
+    def heatPumps = Try.of(() -> csvSystemParticipantSource.getHeatPumps(map(operators), nodeMap, map(types), map(thermalBuses)), SourceException)
+
+    heatPumps.failure
+    heatPumps.exception.get().class == SourceException
+
+    where:
+    nodes               | operators               | types               | thermalBuses              | resultingSize || resultingSet
```
sebastian-peter, line 120: I think it's my mistake, but the `nodes` attribute is not used here. Should be `map(nodes)` in line 111, which you can just integrate into line 120. 

Please also check the tests below

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
```diff
@@ -92,23 +92,39 @@ class CsvSystemParticipantSourceTest extends Specification implements CsvTestDat
...
       new HpInput(sptd.hpInput.uuid, sptd.hpInput.id, OperatorInput.NO_OPERATOR_ASSIGNED, sptd.hpInput.operationTime, sptd.hpInput.node, sptd.hpInput.thermalBus, sptd.hpInput.qCharacteristics, sptd.hpInput.type)
     ]
-    []                  | []                      | []                  | []                        || 0             | []
-    [sptd.hpInput.node] | []                      | []                  | []                        || 0             | []
-    [sptd.hpInput.node] | [sptd.hpInput.operator] | []                  | []                        || 0             | []
-    [sptd.hpInput.node] | [sptd.hpInput.operator] | [sptd.hpInput.type] | []                        || 0             | []
+    [] | [] | [] | []                                                        || 0             | []
+    [sptd.hpInput.node] | []                      | []                  | [] || 0             | []
+    [sptd.hpInput.node] | [sptd.hpInput.operator] | []                  | [] || 0             | []
+    [sptd.hpInput.node] | [sptd.hpInput.operator] | [sptd.hpInput.type] | [] || 0             | []
```
sebastian-peter, line 127: Just a minor point, but the columns are now misaligned here. Please also check below

sebastian-peter (changes requested):
Besides some superficial nodes from my side, this looks good. Thanks!

sebastian-peter (approved):
Thanks, looks great!

PR #988, opened 2024-01-18
Adding log message when using SwitchInputs with `parallelDevices`

### LINKED ISSUE #840: Source cannot read SwitchInput with parallelDevices
This should work as it is a `ConnectorInput` which implements the attribute. Although it always has to be 1 we should maybe be able to read it when in fact it is 1.

```
uuid -> ee77cdb0-8a1d-425e-99c6-5eeec893dc01} are invalid for instance of SwitchInput. 
The following fields (without complex objects e.g. nodes, operators, ...) to be passed to a constructor of 'SwitchInput' are possible (NOT case-sensitive!):
0: [closed, id, uuid]
1: [closed, id, operatesfrom, uuid]
2: [closed, id, operatesuntil, uuid]
3: [closed, id, operatesfrom, operatesuntil, uuid]
```

staudtMarius:
Resolves #840


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Formatting Spotless Groovy import order [#960](https://github.com/ie3-institute/PowerSystemDataModel/issues/960)
-- Implementing missing typical methods in `Try` [#970](https://github.com/ie3-institute/PowerSystemDataModel/issues/970)
+- Implementing missing typical methods in `Try` [#970]()
```
sebastian-peter, line 11: Please repair the link again :)

src/main/java/edu/ie3/datamodel/io/factory/input/SwitchInputFactory.java
```diff
@@ -34,6 +34,19 @@ protected SwitchInput buildModel(
...
+      final int parallelDevices = data.getInt(PARALLEL_DEVICES);
+      log.warn(
+          "Found a `SwitchInput` with the id `{}` that specifies `parallelDevices` with a value of `{}`. Because switches cannot be parallel, the given value is ignored!",
+          id,
+          parallelDevices);
+    } catch (Exception ignored) {
+      // because the field should not be used for switches, we can ignore the exception, that is
+      // thrown, when we
+      // do not find the field
+    }
```
sebastian-peter, line 48: Suggestion from my side:

1. Check with `data.containsKey` if the value exists, this way we can avoid an exception being thrown
2. If the value exists, we use `data.getField` to retrieve the value because it is going to be placed into the log string anyway. We don't need to parse it beforehand. Plus, if the value is not Integer-parseable as we expect, we still get the warning message.

src/main/java/edu/ie3/datamodel/io/factory/input/SwitchInputFactory.java
```diff
@@ -34,6 +34,19 @@ protected SwitchInput buildModel(
       OperatorInput operator,
       OperationTime operationTime) {
     final boolean closed = data.getBoolean(CLOSED);
+
+    try {
+      final int parallelDevices = data.getInt(PARALLEL_DEVICES);
+      log.warn(
+          "Found a `SwitchInput` with the id `{}` that specifies `parallelDevices` with a value of `{}`. Because switches cannot be parallel, the given value is ignored!",
```
sebastian-peter, line 41: I think "switches cannot be parallel" is not true, it's more that we do not specify the amount of parallel lines for switches, as they are supposed to have zero impedance. Maybe you can find a way to improve the warning message a bit.

sebastian-peter (changes requested):

src/main/java/edu/ie3/datamodel/io/factory/input/SwitchInputFactory.java
```diff
@@ -35,16 +35,14 @@ protected SwitchInput buildModel(
...
 
-    try {
-      final int parallelDevices = data.getInt(PARALLEL_DEVICES);
+    if (data.containsKey(PARALLEL_DEVICES)) {
+      String parallelDevices = data.getField(PARALLEL_DEVICES);
+
       log.warn(
-          "Found a `SwitchInput` with the id `{}` that specifies `parallelDevices` with a value of `{}`. Because switches cannot be parallel, the given value is ignored!",
+          "The `SwitchInput` with the id `{}` specifies the unsupported parameter `parallelDevices` with a value of `{}`."
+              + " Because switches are expected to have a given value `1`, the provided value is ignored.",
```
sebastian-peter, line 43: Hmmm, the implicit value is not really 1, is it? I guess it really depends on the context. Maybe something like `SwitchInputs do not need to specify parallelDevices, as its value depends on the electrotechnical context where the switch is embedded.`

sebastian-peter (changes requested):

sebastian-peter (approved):
Just made some tiny changes, thanks for the work!

PR #992, opened 2024-01-24
Pp/#875 fix electricvehicle documentation

pierrepetersmeier:
Has this Issue been processed before? I haven't noticed any significant changes.

PR #993, opened 2024-01-24
!test

pierrepetersmeier:


sebastian-peter:
@t-ober Same question here, besides the one missing attribute, the doc does not seem quite different to the implementation. Maybe we're missing something?

t-ober (changes requested):
image

Okay there is weird stuff going on. When searching for Thermal House I get links to two models, the first one being an outdated earlier version. We should find out what is going on here. Please double check if that happens to you as well.

The same happens for electric vehicle:

image

This is also where the confusion in #992 comes from

sebastian-peter:
@t-ober One is documentation for the input model, the other for the result, right?

t-ober:
> image
> 
> Okay there is weird stuff going on. When searching for Thermal House I get links to two models, the first one being an outdated earlier version. We should find out what is going on here. Please double check if that happens to you as well.
> 
> The same happens for electric vehicle:
> 
> image
> 
> This is also where the confusion in #992 comes from

Okay turns out i am just an idiot. As @sebastian-peter pointed out to me one is the input and one is the result model.

t-ober (approved):

PR #997, opened 2024-01-26
Removing uuid column from time series

### LINKED ISSUE #826: Remove `uuid` as a required column from input and result time series
For most time series, the `time` column is sufficient for uniquely addressing rows. Furthermore the combination of `time` and `coordinate` is a unique key for weather data.
Thus, we do not need (and do not want, due to its size) the redundant column `uuid`.

Documentation also needs to be updated accordingly, of course.

### Affected input data:
- `TimeSeriesEntry` 
  - class can probably be removed
  - affects weather data, coordinate data and "individual time series" (power, power & heat etc.)
- `ResultEntity`
  - can no longer extend `UniqueEntity`
  - affects system participant results, asset result, flex options result, thermal entity results

### Examples

Apparent power time series would have the columns:
```p,q,time```
instead of 
```uuid,p,q,time```

Cosmo weather before:
```uuid,coordinateid,time,diffuseirradiance,directirradiance,temperature,winddirection,windvelocity```
after:
```coordinateid,time,diffuseirradiance,directirradiance,temperature,winddirection,windvelocity```

### Related discussion 
- https://github.com/ie3-institute/simona/discussions/415

cc @t-ober @danielfeismann fyi

### PR DESCRIPTION
**Pull request by staudtMarius**
Resolves #826

### CODE REVIEW

**Review by sebastian-peter — CHANGES_REQUESTED** [review]
Thanks for fighting through this, this definitely looks like a lot more work than I anticipated.

There's just some tiny issues I noticed that can be fixed right away, and for the issues in `CsvDataSource` I would wait for #999 to be merged. We can have a look at possible solutions together then.

File: src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java [hunk 1]
```diff
@@ -342,6 +351,21 @@ protected List<Map<String, String>> csvRowFieldValueMapping(
         .toList();
   }
 
+  public Set<Map<String, String>> checkForDuplicates(
+      Collection<Map<String, String>> rows, String entityDescriptor) {
+    Set<Map<String, String>> rowsSet = new HashSet<>(rows);
```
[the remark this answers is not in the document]
sebastian-peter: Maybe in `EntityData` we need a method to tell us unique fields to check?
So for `UniqueEntity`, this would be `[['uuid'],['id']]` because both uuid and id need to be unique on their own.
And for some timeseries entry of some kind this would maybe be `[['inputModel', 'time']]` or just `[['time']]`, if it's just a time series for a single source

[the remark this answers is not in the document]
staudtMarius: I think I found a good solution for this.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -360,61 +357,95 @@ protected List<Map<String, String>> csvRowFieldValueMapping(
...
+   */
+  protected Try<Void, SourceException> checkUniqueness(
+      String entityName,
+      Collection<Map<String, String>> rows,
+      String fieldName,
+      Set<String> fields) {
+
+    Function<Map<String, String>, String> extractor =
+        fieldsToAttributes ->
+            fields.stream().map(fieldsToAttributes::get).collect(Collectors.joining(""));
```
sebastian-peter, line 415: I think building a String out of all values is not better than calculating a hashcode - it'd really be best to just check unique fields only

sebastian-peter: Let me work out a suggestion...

staudtMarius: This is used if we need the combintion of two fields (e.g. latitude/longitude).

sebastian-peter: Ok, I introduced `List<Set<String>> getUniqueFields()` in `Factory` now (and also realized factories are nowhere near in sight where we need them here, but that's a different problem)

I also understand now how your comparison method works, and it's actually close to what I thought of, as well. What would be even easier (and faster, I think) is just creating sets of values that are supposed to be unique. So, when there are fields uuid and id that are supposed to be unique in combination, you just create a set of the given uuid and id for each row and compare uniqueness amongst all of them.

staudtMarius: So You mean like List.of(Set.of("uuid", "id")) instead of "uuidid"?

sebastian-peter: Maybe passing through the List<Set<String>> uniqueFields down to `buildStreamWithFieldsToAttributesMap` is the way... I'm not quite sure yet

sebastian-peter: > So you mean like List.of(Set.of("uuid", "id")) instead of "uuidid"?

Yes, exactly

sebastian-peter: Where `uuid` and `id` are the respective values, of course. So it'd be something like `Set.of("79b6bcc0-56c7-4e9f-a4d9-e9c7e1482dbd", "entity_name")` for a single entity

sebastian-peter: Or even better, a `Map<String, String>` from field name to value? Should work as well

src/main/java/edu/ie3/datamodel/io/factory/timeseries/TimeSeriesMappingFactory.java
```diff
@@ -38,4 +38,9 @@ protected TimeSeriesMappingSource.MappingEntry buildModel(EntityData data) {
     UUID timeSeries = data.getUUID(TIME_SERIES);
```
sebastian-peter, line 38: Could you finish the adaptation of `TimeSeriesMappingFactory` here to not use UUIDs? Also the new method below

src/main/java/edu/ie3/datamodel/io/factory/timeseries/TimeBasedWeatherValueFactory.java
```diff
@@ -50,4 +49,9 @@ public String getCoordinateIdFieldString() {
    * @return the field name for the date time
    */
   public abstract String getTimeFieldString();
+
+  @Override
+  public List<Set<String>> getUniqueFields() {
```
sebastian-peter, line 54: Same as `TimeSeriesMappingFactory`

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -360,61 +365,95 @@ protected List<Map<String, String>> csvRowFieldValueMapping(
...
+   * @return a try object
+   */
+  protected Try<Void, SourceException> checkUniqueness(
+      String entityName, Collection<Map<String, String>> rows, Set<String> fields) {
+    String fieldName = String.join("-", fields).toUpperCase();
+
+    Function<Map<String, String>, List<String>> extractor =
+        fieldsToAttributes -> fields.stream().map(fieldsToAttributes::get).toList();
+
+    List<List<String>> elements = rows.stream().map(extractor).toList();
```
sebastian-peter, line 417: Reminder for myself to look into this one more time

sebastian-peter (changes requested):

sebastian-peter (commented):
Ok, after looking into duplicate detection more (and also remembering we do something similar in `ValidationUtils` as well), I recognized we have a dilemma that is hard to solve 100%.

One one hand,
- we probably only need duplicate detection for csv source, since databases can do this on their own
- it's convenient to store information about unique field sets with respective factories,

but on the other hand
- currently we use factories only far away from `[Csv]DataSource` since we only work with String maps within data sources
- we do not want to spread validation code all over the project (we already have a "source validation" that checks if all required fields are present, and the regular entity validation)

Given these arguments, I tend to lean more to the side of including duplicate validation with `ValidationUtils` etc., even if that means scrapping same of our work here (`getUniqueFields`, `checkUniqueness`).

So what I envision here is somehow merging `CsvDataSource.checkUniqueness` and `ValidationUtils.checkForDuplicates` and keeping the concept of field sets

src/main/java/edu/ie3/datamodel/models/input/InputEntity.java
```diff
@@ -1,17 +1,11 @@
...
 
-/** Functionless class to describe that all subclasses are input classes */
-public abstract class InputEntity extends UniqueEntity {
-
-  protected InputEntity(UUID uuid) {
-    super(uuid);
-  }
-}
+/** Interface to describe all input entities. */
+public interface InputEntity extends Entity {}
```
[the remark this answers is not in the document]
staudtMarius: This is necessary for the `MappingEntry`, because it is an `InputEntety` that does not have an `UUID`. I changed the previous class `InputEntity` into `UniqueInputEntity` which implements this interface.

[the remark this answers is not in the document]
sebastian-peter: Well yes, but `MappingEntry` and `UniqueInputEntity` can now just extend `Entity`, right?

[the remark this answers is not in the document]
staudtMarius: This interface is used for the `InputDataSink` and the `InputEntityProcessor`. If we remove this interface we have to rewrite some of these classes.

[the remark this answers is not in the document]
sebastian-peter: Got it, thanks.

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
```diff
@@ -303,44 +305,124 @@ protected static void detectMalformedQuantities(
...
+        Try.getExceptions(fieldSets.stream().map(e -> checkUniqueness(entities, e)));
+
+    return Try.ofVoid(
+        !exceptions.isEmpty(),
+        () ->
+            new DuplicateEntitiesException(
+                "The following exception(s) occurred while checking the uniqueness of '"
+                    + entityName
+                    + "' entities: "
+                    + ExceptionUtils.getMessages(exceptions)));
```
sebastian-peter, line 336:
```suggestion
    return entities.stream()
        .findAny()
        .map(
            examplesEntity -> {
              @SuppressWarnings("unchecked")
              Class<E> entityClass = (Class<E>) examplesEntity.getClass();
              String entityClassName = entityClass.getSimpleName();

              List<FieldSetSupplier<E>> fieldSets = getFieldSets(entityClass);

              List<DuplicateEntitiesException> exceptions =
                  Try.getExceptions(fieldSets.stream().map(e -> checkUniqueness(entities, e)));

              return Try.ofVoid(
                  !exceptions.isEmpty(),
                  () ->
                      new DuplicateEntitiesException(
                          "The following exception(s) occurred while checking the uniqueness of '"
                              + entityClassName
                              + "' entities: "
                              + ExceptionUtils.getMessages(exceptions)));
            })
        .orElse(Success.empty());
```

src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java
```diff
@@ -364,16 +354,8 @@ protected List<Map<String, String>> csvRowFieldValueMapping(
...
-   * Checking the uniqueness for a given entity.
-   *
-   * @param entityName name of the entity
-   * @param allRows of the csv file
-   * @param fields list of set of field names
-   * @return a try object
-   */
-  protected Try<Stream<Map<String, String>>, SourceException> checkUniqueness(
-      String entityName, Collection<Map<String, String>> allRows, List<Set<String>> fields) {
+  protected Set<Map<String, String>> checkExactDuplicates(
```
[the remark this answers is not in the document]
staudtMarius: This is for filtering out full (1:1) matching lines.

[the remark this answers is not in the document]
sebastian-peter: I don't think so, to be honest. It's quite costly because you have to calculate hash values of _all_ parameters combined of all rows. And, crucially, if you check for duplicates with your new implementation, that includes duplicate rows as well because they necessarily also have duplicate keys.

src/main/java/edu/ie3/datamodel/utils/validation/ValidationUtils.java
sebastian-peter: This btw should also fix the "bug" detected by sonarqube

src/main/java/edu/ie3/datamodel/utils/validation/GridContainerValidationUtils.java
```diff
@@ -42,10 +41,10 @@ private GridContainerValidationUtils() {
...
     }
 
-    /* sanity check to ensure distinct UUIDs */
-    List<Try<Void, ? extends ValidationException>> exceptions =
-        new ArrayList<>(
-            checkForDuplicates(gridContainer.allEntitiesAsList(), UniqueEntity::getUuid));
+    List<Try<Void, ? extends ValidationException>> exceptions = new ArrayList<>();
+
+    /* sanity check to ensure uniqueness */
+    exceptions.add(checkUniqueness(gridContainer.allEntitiesAsList()));
```
sebastian-peter, line 47: We're checking for duplicates twice: Once here with all entities, then below again with grid elements only, system participants only etc.
I think the problem with this one is that it should actually fail, because the first entity is an `AssetInput`, but there are also entities without id (`GraphicInputs`).

So I think we need to structure validation differently (which also has the benefit of being more user-friendly anyways). My suggestion would be offering various methods for validating asset inputs, result entities, time-based entities etc. (and also a general one for UniqueEntities), and each time the right field set supplier is automatically taken along. 

Sure this produces more methods, but each time it's typesafe and it's clear what is supposed to happen.

sebastian-peter (changes requested):

staudtMarius:
image
I found this method in `SqlDataSource`. This method is currently not used anywhere. Can this method be removed or is there a reason to keep it?

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -205,10 +205,7 @@ private Try<Stream<Map<String, String>>, SourceException> buildStreamWithFieldsT
       // is wanted to avoid a lock on the file), but this causes a closing of the stream as well.
       // As we still want to consume the data at other places, we start a new stream instead of
       // returning the original one
-      Collection<Map<String, String>> allRows =
-          dataSource.csvRowFieldValueMapping(reader, headline);
-
-      return Success.of(dataSource.checkExactDuplicates("Weather", allRows).parallelStream());
+      return Success.of(dataSource.csvRowFieldValueMapping(reader, headline).parallelStream());
```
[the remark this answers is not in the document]
sebastian-peter: I mean the way you implemented, comparing only the unique columns.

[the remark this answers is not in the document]
sebastian-peter: Yes, sorry for the confusion. It shouldn't be a big change, just applying the methods you created, if possible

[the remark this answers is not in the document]
staudtMarius: The question is, do I apply the method to the build data or to the raw strings?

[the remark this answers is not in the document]
staudtMarius: You mean applying to the build data?

src/main/java/edu/ie3/datamodel/utils/validation/UniquenessValidationUtils.java
```diff
@@ -0,0 +1,177 @@
...
+  // default field set supplier
+  protected static final FieldSetSupplier<UniqueEntity> uuidFieldSupplier =
+      entity -> Set.of(entity.getUuid());
+  protected static final FieldSetSupplier<AssetInput> idFieldSupplier = e -> Set.of(e.getId());
+  protected static final FieldSetSupplier<ResultEntity> resultFieldSupplier =
+      entity -> Set.of(entity.getTime(), entity.getInputModel());
+  protected static final FieldSetSupplier<MappingEntry> mappingFieldSupplier =
+      entity -> Set.of(entity.participant());
+  protected static final FieldSetSupplier<TimeBasedValue<WeatherValue>>
+      timeBasedValueFieldSupplier =
```
sebastian-peter, line 34: One more point: This should probably be `weatherValueFieldSupplier`, right?

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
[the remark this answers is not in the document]
sebastian-peter: Yes, so in `CsvWeatherSource` this would happen somewhere towards the end of `readWeatherTimeSeries`. Or are there arguments against including the uniqueness validation?

src/main/java/edu/ie3/datamodel/utils/validation/UniquenessValidationUtils.java
staudtMarius: Yes, that is a better name.

src/main/java/edu/ie3/datamodel/models/input/IdCoordinatePair.java
```diff
@@ -0,0 +1,45 @@
...
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.models.input;
+
+import edu.ie3.util.geo.GeoUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.locationtech.jts.geom.Point;
+
+public class IdCoordinatePair extends ImmutablePair<Integer, Point> implements InputEntity {
```
sebastian-peter, line 12: Hmm it's not a pair anymore, right? So could just be named `IdCoordinateInput`?

src/main/java/edu/ie3/datamodel/models/input/IdCoordinatePair.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Create a new pair instance.
+   *
+   * @param left the left value, may be null
+   * @param right the right value, may be null
+   */
+  private IdCoordinatePair(Integer left, Point right) {
```
sebastian-peter, line 19: `left` and `right` are not really fitting names anymore I think...

src/main/java/edu/ie3/datamodel/models/input/IdCoordinatePair.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Creates an {@link IdCoordinatePair}.
+   *
+   * @param id of the pair
+   * @param lat of the pair
+   * @param lon of the pair
+   * @return a new {@link IdCoordinatePair}
+   */
+  public static IdCoordinatePair of(Integer id, double lat, double lon) {
```
sebastian-peter, line 31: This could just be a regular constructor `public IdCoordinatePair(Integer id, ...)`

src/main/java/edu/ie3/datamodel/models/input/IdCoordinatePair.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  /**
+   * Creates an {@link IdCoordinatePair}.
+   *
+   * @param id of the pair
+   * @param point of the pair
+   * @return a new {@link IdCoordinatePair}
+   */
+  public static IdCoordinatePair of(Integer id, Point point) {
```
sebastian-peter, line 42: Same here

docs/readthedocs/io/ValidationUtils.md
```diff
@@ -49,15 +65,14 @@ The ValidationUtils include validation checks for...
         - PvInput
         - StorageInput
         - WecInput
-        - (missing: EvcsInput)
+        - EvcsInput (also checks EvcsTypeInput/ChargingPointType)
```
sebastian-peter, line 68: `EvcsTypeInput` still doesn't exist...

src/main/java/edu/ie3/datamodel/models/input/IdCoordinatePair.java
sebastian-peter, line 1: We could've also loosened the requirements on types checked by `UniquenessValidationUtils`, but overall this is probably the nicer solution. So, thanks!

sebastian-peter (approved):
@staudtMarius This is it! I only made tiny adaptations at the end. Thanks a lot for your work, this is a big step towards a more efficient PSDM.

PR #1002, opened 2024-02-02
Fix equals behavior of `ConnectorResult`

### LINKED ISSUE #1001: Fix `equals`behavior of `ConnectorResult`
Currently we are not comparing their respective input models, which leads to equal results of different models being dropped in sets. And it seems like the equals method is used to determine whether objects are duplicates in sets.

"Even though the hashCode of two objects may be different, what ultimately determines whether two objects are considered duplicates in a set is the equals method. The hashCode method is used primarily for optimizing lookups in hashed data structures, such as HashSet. If two objects are equal according to their equals method, they must also have the same hashCode. However, having the same hashCode does not necessarily mean two objects are equal."

t-ober:
Resolves #1001


sebastian-peter (approved):
LGTM, good catch

PR #1012, opened 2024-02-09
Introducing scaling function in copy builders

### LINKED ISSUE #1011: Implement scaling function in system participant inputs
The current method of scaling system participants in simona is not sufficient, as storage capacities etc. are not scaled, while output power is.
A convenient way of scaling participants would be to offer a method `SystemParticipantInput scale(Double factor)` that returns a scaled copy of itself.

sebastian-peter:
Closes #1011

Also implementing a bunch of missing copy builders

### DIFF
--- a/src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
@@ -1,0 +1,49 @@
+/*
+ * © 2024. TU Dortmund University,
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+ */
+package edu.ie3.datamodel.models.input.system.type.chargingpoint
+
+import edu.ie3.datamodel.models.ElectricCurrentType
+import edu.ie3.datamodel.models.StandardUnits
+import spock.lang.Specification
+import tech.units.indriya.quantity.Quantities
+
+class ChargingPointTypeTest extends Specification {
+
+  def "A ChargingPointType copy method should work as expected"() {
+    given:
+    def cpt = ChargingPointTypeUtils.HouseholdSocket
+
+    when:
+    def alteredUnit = cpt.copy()
+        .setElectricCurrentType(ElectricCurrentType.DC)
+        .setsRated(Quantities.getQuantity(2.4, StandardUnits.S_RATED))
+        .build()
+
+    then:
+    alteredUnit.with {
+      assert id == cpt.id
+      assert sRated == Quantities.getQuantity(2.4, StandardUnits.S_RATED)
+      assert electricCurrentType == ElectricCurrentType.DC
+      assert synonymousIds == cpt.synonymousIds
+    }
+  }
+
+  def "Scaling a ChargingPointType via builder should work as expected"() {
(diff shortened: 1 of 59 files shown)


sebastian-peter:
@reviewers: I'd be happy if you took a closer look into what parameters I do scale and I don't scale here and whether you agree with me!

danielfeismann:
A more general question. How should we deal with capex? I guess we should scale it for all models as well even this does not represent economy of scale...

src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
```diff
@@ -0,0 +1,49 @@
...
+class ChargingPointTypeTest extends Specification {
+
+  def "A ChargingPointType copy method should work as expected"() {
+    given:
+    def cpt = ChargingPointTypeUtils.HouseholdSocket
+
+    when:
+    def alteredUnit = cpt.copy()
+        .setElectricCurrentType(ElectricCurrentType.DC)
+        .setsRated(Quantities.getQuantity(2.4, StandardUnits.S_RATED))
```
danielfeismann, line 22: This should be 2.3, or?

src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
```diff
@@ -0,0 +1,49 @@
...
+        .build()
+
+    then:
+    alteredUnit.with {
+      assert id == cpt.id
+      assert sRated == Quantities.getQuantity(2.4, StandardUnits.S_RATED)
```
danielfeismann, line 28: same here

danielfeismann (changes requested):
This looks very good to me. Just the capex question and one minor thing.

src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
sebastian-peter: The household socket has 2.3 kVA originally (as per `ChargingPointTypeUtils`), and to check whether the copy builder works, we change that value to 2.4 kVA and test whether that value has been applied to the new `ChargingPointType`.

https://github.com/ie3-institute/PowerSystemDataModel/blob/667680f35d97170f0c83563c0323bcaf72a07bae/src/main/java/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeUtils.java#L40-L45

sebastian-peter:
> A more general question. How should we deal with capex? I guess we should scale it for all models as well even this does not represent economy of scale...

I think we should, just to approximate the change. If one looked closely, `opex` could also reduce while scaling up, but for those precise model changes, one should just properly re-model their scenario.

src/test/groovy/edu/ie3/datamodel/models/input/system/type/chargingpoint/ChargingPointTypeTest.groovy
danielfeismann: True. It didn't get the setsRated..

danielfeismann (approved):
LGTM, thanks for quickly including capex as well.

PR #1020, opened 2024-02-26
Implementing `CsvThermalGridSource`.

### LINKED ISSUE #1009: Implement `CsvThermalGridSource`
One can take the [SIMONA implementation](https://github.com/ie3-institute/simona/blob/dev/src/main/scala/edu/ie3/simona/io/grid/CsvGridSource.scala) as an inspiration.

Also, a recent feature in `ThermalSource` can be reverted. `getThermalBuses()`, `getThermalStorages()` and `getThermalHouses()` return maps by model uuid, but this is actually not needed, since they are sorted by bus uuid later.

staudtMarius:
Resolves #1009


docs/readthedocs/models/models.md
sebastian-peter, line 1: Thanks for cleaning up the doc structure!

sebastian-peter (changes requested):
Ok I think I was wrong about rolling back the thermal entity structures to sets. Maps do make sense here because one has to match the thermal entities to HPs and CHPs. Sorry. Could you roll back these specific changes?

src/main/java/edu/ie3/datamodel/utils/CollectionUtils.java
sebastian-peter, line 1: To be honest I don't know if we should introduce this utils class if it's (at most) used at two places. Both `toMap` methods are not used anymore and `groupBy` is quite specific if you can just use `collect(Collectors.groupingBy(keyExtractor))` most of the time (if a value lists are fine).

In fact, if one really needs value sets like we do in `CsvThermalGridSource`, the same result can be achieved by using a different `Collectors.groupingBy` method:
```java
      Map<ThermalBusInput, Set<ThermalHouseInput>> houseInputs =
          houses.getOrThrow().stream()
              .collect(Collectors.groupingBy(ThermalUnitInput::getThermalBus, Collectors.toSet()));
```

src/main/java/edu/ie3/datamodel/utils/Try.java
```diff
@@ -593,7 +594,7 @@ public int hashCode() {
    */
   @FunctionalInterface
   public interface TrySupplier<T, E extends Exception> {
-    T get() throws E;
+    T get() throws E, SourceException;
```
sebastian-peter, line 597: Where/why is this needed?

sebastian-peter (changes requested):
I think we can do this with less code...

sebastian-peter (approved):
Looks great, thanks!

PR #1024, opened 2024-02-26
Fix CsvGraphicSource testing

### LINKED ISSUE #1022: Fix CsvGraphicSource testing when processing invalid data
CsvGraphicSource should fail if there is no entity for a graphic input.

jo-bao:
Resolves #1022


sebastian-peter (approved):
Nice catch! The test definitely makes more sense like this.

PR #1038, opened 2024-03-06
Fixing equals in `ResultEntity` and `TimeSeriesEntry`

### LINKED ISSUE #1037: Fix `equals` of `ResultEntity` and `TimeSeriesEntry`
Regression of #997

Both classes extend `Entity` which does not provide an implementation of `equals`. Thus, the implementation of `Object` is used when calling `super.equals`, which is not what we want

sebastian-peter:
Resolves #1037


danielfeismann (approved):
Thanks for fixing this

PR #1044, opened 2024-03-08
Abstract commonly used functionality from `EntitySource`.

### LINKED ISSUE #975: Make `EntitySource` completely static
After refinements related to #969, the `DataSource` that is required for instanciating `EntitySource` is effectively only used in `EntitySource#buildEntityData`. Since all other methods are static or build upon this method, one can get confused easily here: When invoking one of these methods, does the output result from the input parameters, the field `EntitySource.dataSource`, or both?

Thus, I'm suggesting to make `EntitySource` completely static. Where data is needed from `DataSource`, either this object or a list of field-to-attribute sets can be passed as a parameter.

One can also discuss whether to also apply this transformation to subclasses of `EntitySource`, although this might be a bit more complex.

### LINKED ISSUE #981: Abstract commonly used functionality from `EntitySource`
Especially with #926 it became apparent that some sources re-implement functionality that already exists in `EntitySource`.
This functionality includes (but might not be limited to) methods 

```java
public abstract void validate() throws ValidationException;
<C extends UniqueEntity> Try<Void, ValidationException> validate(Class<? extends C>, SourceValidator<C>)
```
, which could be reused by classes
- `*WeatherSource` (`CsvWeatherSource` might be a bit more difficult)
- `*IdCoordinateSource`
- `*TimeSeriesSource`

where `*` can be multiple of `Sql`, `Csv`, `Couchbase`, `InfluxDb`.

Probably the methods above have to be adapted to abstract the functionality properly.

In the end, we'll probably have `AssetEntitySource`, which takes over most of the current `EntitySource` implementation, and a new `EntitySource` which will be extended by `AssetEntitySource` and by the classes mentioned above.

staudtMarius:
Resolves #975
Resolves #981

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
@@ -99,7 +99,7 @@ public class GraphicSource extends AssetEntitySource {
 
     if (!exceptions.isEmpty()) {
       throw new GraphicSourceException(
-          exceptions.size() + "error(s) occurred while initializing graphic elements. ",
+          exceptions.size() + " error(s) occurred while initializing graphic elements. ",
           exceptions);
     } else {
       // if everything is fine, return a GraphicElements instance
@@ -120,13 +120,12 @@ public class GraphicSource extends AssetEntitySource {
 
   public Set<NodeGraphicInput> getNodeGraphicInput(Map<UUID, NodeInput> nodes)
       throws SourceException {
-    return toSet(
-        getEntities(
+    return getEntities(
             NodeGraphicInput.class,
             dataSource,
             nodeGraphicInputFactory,
-            data ->
-                enrich(data, buildEnrichment(data, NODE, nodes), NodeGraphicInputEntityData::new)));
+            enrich(NODE, nodes, NodeGraphicInputEntityData::new))
+        .collect(toSet());
   }
 
   /**
@@ -143,13 +142,11 @@ public class GraphicSource extends AssetEntitySource {
 
   public Set<LineGraphicInput> getLineGraphicInput(Map<UUID, LineInput> lines)
       throws SourceException {
-    return toSet(
-        getEntities(
+    return getEntities(
             LineGraphicInput.class,
             dataSource,
             lineGraphicInputFactory,
-            data ->
-                enrich(
-                    data, buildEnrichment(data, "line", lines), LineGraphicInputEntityData::new)));
+            enrich("line", lines, LineGraphicInputEntityData::new))
+        .collect(toSet());
   }
 }
--- a/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
@@ -131,7 +131,7 @@ public class GraphicSource extends AssetEntitySource {
   /**
    * If the set of {@link LineInput} entities is not exhaustive for all available {@link
    * LineGraphicInput} entities or if an error during the building process occurs a {@link
-   * SourceException} is thrown, else all entities that have been able to be build are returned.
+   * SourceException} is thrown, else all entities that have been able to be built are returned.
    */
   public Set<LineGraphicInput> getLineGraphicInput() throws SourceException {
     Map<UUID, OperatorInput> operators = typeSource.getOperators();
--- a/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
@@ -22,8 +22,10 @@ import edu.ie3.datamodel.models.input.graphics.GraphicInput;
 import edu.ie3.datamodel.models.input.graphics.LineGraphicInput;
 import edu.ie3.datamodel.models.input.graphics.NodeGraphicInput;
 import edu.ie3.datamodel.utils.Try;
-import edu.ie3.datamodel.utils.Try.*;
-import java.util.*;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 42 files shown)


src/main/java/edu/ie3/datamodel/io/source/GraphicSource.java
```diff
@@ -118,15 +120,18 @@ public Set<NodeGraphicInput> getNodeGraphicInput() throws SourceException {
...
+            nodeGraphicInputFactory,
+            enrich(NODE, nodes, NodeGraphicInputEntityData::new))
+        .collect(toSet());
   }
 
   /**
    * If the set of {@link LineInput} entities is not exhaustive for all available {@link
    * LineGraphicInput} entities or if an error during the building process occurs a {@link
-   * SourceException} is thrown, else all entities that have been able to be built are returned.
+   * SourceException} is thrown, else all entities that have been able to be build are returned.
```
sebastian-peter, line 134: This was correct before, we need the past participle of "build" here

src/main/java/edu/ie3/datamodel/io/source/SystemParticipantSource.java
```diff
@@ -680,154 +724,29 @@ public Set<HpInput> getHeatPumps(
...
-                nodeAssetInputEntityDataTry.flatMap(
-                    nodeAssetInputEntityData ->
-                        optionallyEnrichEntityData(
-                            nodeAssetInputEntityData,
-                            SystemParticipantInputEntityFactory.EM,
-                            emUnits,
-                            null,
-                            SystemParticipantEntityData::new)));
+  private static <T extends SystemParticipantTypeInput, D extends SystemParticipantEntityData>
+      TryFunction<D, SystemParticipantTypedEntityData<T>> enrich(Map<UUID, T> types) {
```
sebastian-peter, line 747: Same here as above, please rename function

src/main/java/edu/ie3/datamodel/io/source/csv/CsvIdCoordinateSource.java
```diff
@@ -53,11 +53,21 @@ public CsvIdCoordinateSource(IdCoordinateFactory factory, CsvDataSource dataSour
     this.factory = factory;
     this.dataSource = dataSource;
 
+    // validating
+    Try.ofVoid(this::validate, ValidationException.class)
+        .transformF(SourceException::new)
+        .getOrThrow();
```
sebastian-peter, line 59: So here we start validating within the constructor, but in weather sources you remove the validation from constructor. Is there a reason for that?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlIdCoordinateSource.java
```diff
@@ -58,19 +56,6 @@ public SqlIdCoordinateSource(
...
-            fieldsOpt ->
-                fieldsOpt
-                    .map(
-                        fields ->
-                            factory
-                                .validate(fields, IdCoordinateInput.class)
-                                .transformF(SourceException::new))
-                    .orElse(Try.Success.empty()))
-        .getOrThrow();
-
```
sebastian-peter, line 73: ... adding validation to the equivalent csv source, removing it here?

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
```diff
@@ -159,7 +157,6 @@ class CsvSystemParticipantSourceTest extends Specification implements CsvTestDat
 
     where:
     operators               | types               | thermalBuses              || resultingSize | resultingSet
-    []                      | [sptd.hpInput.type] | [sptd.hpInput.thermalBus] || 1             | [new HpInput(sptd.hpInput.uuid, sptd.hpInput.id, OperatorInput.NO_OPERATOR_ASSIGNED, sptd.hpInput.operationTime, sptd.hpInput.node, sptd.hpInput.thermalBus, sptd.hpInput.qCharacteristics, sptd.emInput, sptd.hpInput.type)]
```
sebastian-peter, line 162: I don't agree with just removing the successful cases completely. What is the problem here?

sebastian-peter (changes requested):
I like your implementation of the issue a lot. This is more than I expected! I just have some minor remarks.

Additionally, please also adapt `EntitySourceClassDiagram.puml`. If you think that it makes sense, you can leave out protected and private methods there from now on.

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
staudtMarius: In theses test we are explicitly testing cases that are leading to failures.

sebastian-peter: Ok, I see. `resultingSize` and `resultingSet` is not used anymore and can be removed then

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvSystemParticipantSourceTest.groovy
```diff
@@ -283,7 +276,6 @@ class CsvSystemParticipantSourceTest extends Specification implements CsvTestDat
 
     where:
     operators               | types               || resultingSet
-    []                      | [sptd.bmInput.type] || [new BmInput(sptd.bmInput.uuid, sptd.bmInput.id, OperatorInput.NO_OPERATOR_ASSIGNED, sptd.bmInput.operationTime, sptd.bmInput.node, sptd.bmInput.qCharacteristics, sptd.emInput, sptd.bmInput.type, sptd.bmInput.marketReaction, sptd.bmInput.costControlled, sptd.bmInput.feedInTariff)]
     [sptd.bmInput.operator] | []                  || []
     []                      | []                  || []
     []                      | []                  || []
```
sebastian-peter, line 281: Also this is just duplicated...

src/main/java/edu/ie3/datamodel/utils/Try.java
sebastian-peter, line 1: Ok, something I just noticed. Could you take care of testing the new methods in `TryTest`? Thanks!

sebastian-peter (changes requested):

sebastian-peter (approved):
Thanks!

PR #1062, opened 2024-04-09
Improving the usage of streams on SQL fetches.

### LINKED ISSUE #827: Use Java streams on SQL fetches effectively
Once #773 is merged:

The call to `stream()` here does not make much sense if one thinks of why we want to use it in the first place. We fetch all rows for a query first, then apply a stream on it to "lazily" handle the rows one by one. 
We definitely should not handle this in this PR though, I think. I found information on how to do it though: https://stackoverflow.com/a/64391247

_Originally posted by @sebastian-peter in https://github.com/ie3-institute/PowerSystemDataModel/pull/773#discussion_r1205376824_

staudtMarius:
Resolves #827


danielfeismann (dismissed):
LGTM, but maybe Sebastian would like to have a look as well?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlDataSource.java
```diff
@@ -177,11 +176,15 @@ protected Stream<Map<String, String>> buildStreamByTableName(String tableName) {
...
+    try {
+      PreparedStatement ps = connector.getConnection().prepareStatement(query);
       addParams.addParams(ps);
 
-      ResultSet resultSet = ps.executeQuery();
-      return connector.extractFieldMaps(resultSet).stream();
+      Stream<Map<String, String>> stream = connector.toStream(ps, 1000);
+
+      // don't work with `try with resource`, but is necessary to get results
+      ps.closeOnCompletion();
```
sebastian-peter, line 186: 1. I think it'd make more sense to call this before the stream is created
2. Checking the JavaDoc of this method, I think this only works if the ResultSet is closed (which then causes the Statement to be closed).

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -115,21 +121,47 @@ public void shutdown() {
...
+        true);
+  }
+
+  /**
+   * Returns an {@link Iterator} for the given {@link ResultSet}.
+   *
+   * @param rs given result set
+   * @return an iterator
+   */
+  public Iterator<Map<String, String>> getSqlIterator(ResultSet rs) {
```
sebastian-peter, line 148: Looking at the StackOverflow answer that I linked in the issue, isn't the purpose of this iterator to close the resources once we've iterated through the results?

sebastian-peter (changes requested):

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -115,21 +121,56 @@ public void shutdown() {
...
+    return new Iterator<>() {
+      @Override
+      public boolean hasNext() {
+        try {
+          return rs.next();
+        } catch (SQLException e) {
+          log.error("Exception at extracting next ResultSet: ", e);
+          return false;
```
sebastian-peter, line 165: I thnk we should also close the ResultSet on error

src/main/java/edu/ie3/datamodel/io/connectors/SqlConnector.java
```diff
@@ -115,21 +121,56 @@ public void shutdown() {
...
+        }
       }
-    } catch (SQLException e) {
-      log.error("Exception at extracting ResultSet: ", e);
-    }
-    return fieldMaps;
+
+      @Override
+      public Map<String, String> next() {
+        return extractFieldMap(rs);
```
sebastian-peter, line 171: Also here. Maybe `extractFieldMap` should not catch the SQLException, but we handle it inside the `next()` method.

sebastian-peter (changes requested):

sebastian-peter (changes requested):
Looks good for now and it seems to work. One more request from my side

sebastian-peter (approved):
Did some last tinkering

PR #1064, opened 2024-04-09
Enhancing `VoltageLevel` with `equals` method.

### LINKED ISSUE #1063: Enhance `VoltageLevel` with `equals` method
This will fix this [issue](https://github.com/ie3-institute/OSMoGrid/issues/395).

staudtMarius:
Resolves #1063


sebastian-peter (changes requested):
Please also provide an implementation for a `hashcode` method, as equals and hashcode should always be provided in combination. Intellij offers some nice auto-generated implementations.

staudtMarius:
> Please also provide an implementation for a `hashcode` method, as equals and hashcode should always be provided in combination. Intellij offers some nice auto-generated implementations.

Now there are also added.

sebastian-peter (approved):
I think it's fine if we do not compare some fields like ids here. However, I fixed the equals method of `CommonVoltageLevel`, which did not call the equals method of the super class before.

PR #1085, opened 2024-05-15
Fixing bug that prevents `MappingEntries` from getting processed.

### LINKED ISSUE #1084: TimeSeriesSource MappingEntries cannot be processed
Because the `TimeSeriesSource.MappingEntry` is a `record` and does not have dedicated `Getter` methods, the `InputEntityProcessor` creates an empty map for those entities.

staudtMarius:
Resolves #1084


danielfeismann (approved):
LGTM

Issue #994, opened 2024-01-24
Fix project being built twice in CI


sebastian-peter:
Looks like PSDM is built twice, once for tests and once for sonarqube. This should not be necessary and SIMONA works in CI without it.

![grafik](https://github.com/ie3-institute/PowerSystemDataModel/assets/14994800/023f0f85-c5d2-4918-8383-f5c26ab199f3)

PR #1105, opened 2024-06-24
Fixing project being build twice in CI

staudtMarius:
Reolves #994


sebastian-peter (approved):
Ok let's just see if this breaks something.

PR #1135, opened 2024-08-09
Adding sources for load profile time series.

### LINKED ISSUE #1106: Transfer LoadProfile parsing code from SIMONA?
`LoadProfileInput`, `LoadProfileEntry` seem to be unused in SIMONA, which has its own implementation instead. Check whether it makes sense to transfer the reading and handling of load profile parameters into PSDM, which seems to be more suited for that.

staudtMarius:
Resolves #1106 

Note MS: Add information to readthedocs.


src/main/java/edu/ie3/datamodel/io/source/sql/SqlLoadProfileSource.java
```diff
@@ -0,0 +1,205 @@
...
+  @Override
+  public LoadProfileTimeSeries<V> getTimeSeries() {
+    Set<LoadProfileEntry<V>> entries = getEntries(queryFull, ps -> {});
+    return entryFactory.build(metaInformation, entries);
+  }
+
+  @Override
+  public List<ZonedDateTime> getTimeKeysAfter(ZonedDateTime time) {
+    return List.of(time.plusMinutes(15));
+  }
```
danielfeismann, line 112: This would only work correct, when the time series are always in quarter hours steps, right? Should this get adapted to (optional) provide the step size and only take 15 minutes if none is provided?

danielfeismann (changes requested):

PR #1145, opened 2024-08-27
Handling no received data from WeatherSource

### LINKED ISSUE #554: Explicitly handle not receiving any weather data from the weather source
Currently we get a division by zero error later down the road when not receiving any weather values from the weather Source. We should throw an exception before explicitly stating that there we did not find any weather data!

PhilippSchmelter:
Resolves #554 
Added an explicit Exception for handling no received data from WeatherSource to prevent the pipeline moving on without data.


src/main/java/edu/ie3/datamodel/exceptions/NoWeatherDataException.java
sebastian-peter, line 1: `NoDataException` and year 2024

PhilippSchmelter:
Handling CSV-Source done
+ all getWeather() which returned optional
//TODO
REST of
SQL
Influx
Couchbase

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -102,21 +103,29 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
+          coordinates.stream()
+              .map(idCoordinateSource::getId)
+              .flatMap(Optional::stream)
+              .collect(Collectors.toSet());
+      if (coordinateIds.isEmpty()) {
+        log.warn("Unable to match coordinates to coordinate ID");
+        throw new NoDataException("No coordinates found");
+      }
+    } catch (NoDataException e) {
+      return null;
```
danielfeismann, line 128: Please check for the SonarQube hints for returning an empty map

danielfeismann (commented):
Thanks for tackling this issue. Just some questions from my side. 

Note to self: Still need to check tests.

PhilippSchmelter:
I still need to write some tests I guess. Checked SonarQube for coverage. There are uncovered Conditions
@sebastian-peter @danielfeismann

danielfeismann:
@PhilippSchmelter are you plan to add some tests here?

CHANGELOG.md
```diff
@@ -88,6 +89,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added load profiles sources [#1106](https://github.com/ie3-institute/PowerSystemDataModel/issues/1106)
 - Add `v2gSupport` parameter to documentation of `EvcsModel` [#1278](https://github.com/ie3-institute/PowerSystemDataModel/issues/1278)
 
+
```
danielfeismann: Please remove this empty line here

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @version 0.1
+ * @since 04.09.24
+ */
+public class NoDataException extends Exception {
+
+  public NoDataException(final String message) {
+    super(message);
+  }
+}
```
danielfeismann, line 21: Can't we use some existing exception type here? NullPointerException may not hundred percent correct but matches it almost, right?

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -138,32 +140,41 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
               new IndividualTimeSeries<>(weatherInputs);
           coordinateToTimeSeries.put(coordinate, weatherTimeSeries);
         }
-      } else logger.warn("Unable to match coordinate {} to a coordinate ID", coordinate);
+      } else {
+        logger.error("Unable to match coordinate {} to a coordinate ID", coordinate);
+        throw new NoDataException("No data found");
```
danielfeismann: It would be great to provide the error message within the exception here.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -138,32 +140,41 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
   @Override
-  public Optional<TimeBasedValue<WeatherValue>> getWeather(ZonedDateTime date, Point coordinate) {
+  public TimeBasedValue<WeatherValue> getWeather(ZonedDateTime date, Point coordinate)
+      throws NoDataException {
     Optional<Integer> coordinateId = idCoordinateSource.getId(coordinate);
     if (coordinateId.isEmpty()) {
-      logger.warn("Unable to match coordinate {} to a coordinate ID", coordinate);
-      return Optional.empty();
+      logger.error("Unable to match coordinate {} to a coordinate ID", coordinate);
+      throw new NoDataException("No coordinate ID found for the given point.");
```
danielfeismann: same here, providing the information which id  or coordinate failed helps a lot

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -138,32 +140,41 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
-      return toTimeBasedWeatherValue(jsonWeatherInput);
+      return toTimeBasedWeatherValue(jsonWeatherInput)
+          .orElseThrow(
+              () ->
+                  new NoDataException(
+                      "No valid weather data found for the given date and coordinate."));
     } catch (DecodingFailureException ex) {
       logger.error("Decoding to TimeBasedWeatherValue failed!", ex);
-      return Optional.empty();
+      throw new NoDataException("Failed to decode weather data.");
```
danielfeismann: Should we add the DecondingFailureException message here again?

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -97,25 +104,43 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
-    Optional<Integer> coordinateId = idCoordinateSource.getId(coordinate);
-    if (coordinateId.isEmpty()) {
-      return Optional.empty();
+  public TimeBasedValue<WeatherValue> getWeather(ZonedDateTime date, Point coordinate)
+      throws NoDataException {
+    Optional<Integer> coordinateId;
+    try {
+      coordinateId = idCoordinateSource.getId(coordinate);
+      if (coordinateId.isEmpty()) {
+        throw new NoDataException("No coordinate ID found for the given point.");
```
danielfeismann: May we give the point here as well?

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -97,25 +104,43 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
     try (InfluxDB session = connector.getSession()) {
       String query = createQueryStringForCoordinateAndTime(date, coordinateId.get());
       QueryResult queryResult = session.query(new Query(query));
-      return filterEmptyOptionals(optTimeBasedValueStream(queryResult)).findFirst();
+      return filterEmptyOptionals(optTimeBasedValueStream(queryResult))
+          .findFirst()
+          .orElseThrow(
+              () ->
+                  new NoDataException(
+                      "No weather data available for the given date and coordinate."));
```
danielfeismann: same here

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
danielfeismann: if so, please check also for sql and cs weather source

danielfeismann (changes requested):
Thanks for tackling this. Some points from my side

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
PhilippSchmelter: Technically we defineteley could use the NullPointerException but I would leave it as it is for semantic clarity. This Exception explicitly communicates the specific problem domain. I think I already talked about this with sebastian, not sure tho.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
PhilippSchmelter: I've made some changes. Let me know if this is somewhat you imagined.

PhilippSchmelter:
> Thanks for tackling this. Some points from my side

I hope I was able to fix the right things or what you meant! Let me know if its good like that.

PhilippSchmelter:
@danielfeismann 
I think now this Issue should be covered :D
Feel free to check the changes and let me know if something doesnt fit!

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Enhance `TimeSeriesSource` with method to retrieve all time keys after a given key [#543](https://github.com/ie3-institute/PowerSystemDataModel/issues/543)
 - Enhance `WeatherSource` with method to retrieve all time keys after a given key [#572](https://github.com/ie3-institute/PowerSystemDataModel/issues/572)
+- Added explicit handling for cases where no weather data is received from any source [#554](https://github.com/ie3-institute/PowerSystemDataModel/issues/554)
```
danielfeismann: Please move this to the end of this subsection

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
```diff
@@ -0,0 +1,24 @@
...
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.exceptions;
+
+/**
+ * Exception that should be used whenever no data is received{@link
+ * edu.ie3.datamodel.io.source.DataSource}
+ *
+ * @version 0.1
+ * @since 04.09.24
```
danielfeismann: Please remove this, we started this somewhen but not really use this and do not update it when changing parts of methods.

src/test/groovy/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSourceIconIT.groovy
```diff
@@ -118,31 +119,42 @@ class InfluxDbWeatherSourceIconIT extends Specification implements WeatherSource
...
+    when: "requesting weather for mixed valid and invalid coordinates"
     def coordinatesToTimeSeries = source.getWeather(timeInterval, [
       validCoordinate,
       invalidCoordinate
     ])
 
-    then:
-    coordinateAtDate == Optional.empty()
-    equalsIgnoreUUID(coordinateInInterval, emptyTimeSeries)
+    then: "only valid coordinate data is returned"
```
danielfeismann: Do we want this? I think if we would throw an exception when requesting some invalid coordinate, then it should still be thrown if it comes along with valid ones.

src/test/groovy/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSourceCosmoIT.groovy
```diff
@@ -125,31 +126,42 @@ class InfluxDbWeatherSourceCosmoIT extends Specification implements TestContaine
...
+
+    when: "requesting weather for an invalid coordinate in a time interval"
+    source.getWeather(timeInterval, invalidCoordinate)
+
+    then: "NoDataException is thrown"
+    def ex2 = thrown(NoDataException)
+    ex2.message.contains("No data for given coordinates")
+    ex2.message.contains(invalidCoordinate.toString())
+
+    when: "requesting weather for mixed valid and invalid coordinates"
```
danielfeismann, line 151: see this case above

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -97,25 +103,37 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
               optTimeBasedValueStream(queryResult);
           Set<TimeBasedValue<WeatherValue>> timeBasedValues =
               filterEmptyOptionals(optValues).collect(Collectors.toSet());
-          IndividualTimeSeries<WeatherValue> timeSeries =
-              new IndividualTimeSeries<>(timeBasedValues);
-          coordinateToTimeSeries.put(entry.getKey(), timeSeries);
+          if (!timeBasedValues.isEmpty()) {
```
danielfeismann: Should it also lead to an exception in case timeBasedValues is empty, since one could assume we would expect weather data when asking for it and those aren't there? Not sure about this, I might have to rethink on this.

danielfeismann: Also, testing this would also be great.

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
danielfeismann: Thanks for pointing that out again, I also reendevoured above that this was in one of my previous comments :)

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
danielfeismann: Thanks for including this!

danielfeismann (changes requested):
Thank you for including my points. I do have another one or two points which are a bit unclear to me

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
PhilippSchmelter: Alright!

src/test/groovy/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSourceIconIT.groovy
PhilippSchmelter: I dont know if Im wrong at this point, but if I remember and understood correctly the approaches for handling this were not consistent through the various WeatherSources. Still have to check that, yesterday was kind of a long coding session :D
But I will change everything to strict failure if no or damaged data is found. I think this would be the safest approach and saving the user time by following the fail fast philosophy.

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
PhilippSchmelter: I tested throwing an exception when timeBasedValues is empty, but that broke existing behavior and tests. For now we keep it simple: invalid coordinates throw a NoDataException, but valid coordinates with no data just return an empty series. Or do we really want it that strict? If so we could use my last commit which I reverted again.

CHANGELOG.md
sebastian-peter: Reminder to move this to the `Unreleased/Snapshot` section

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -138,32 +151,52 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
   @Override
-  public Optional<TimeBasedValue<WeatherValue>> getWeather(ZonedDateTime date, Point coordinate) {
+  public TimeBasedValue<WeatherValue> getWeather(ZonedDateTime date, Point coordinate)
+      throws NoDataException {
     Optional<Integer> coordinateId = idCoordinateSource.getId(coordinate);
     if (coordinateId.isEmpty()) {
-      logger.warn("Unable to match coordinate {} to a coordinate ID", coordinate);
-      return Optional.empty();
+      logger.error("Unable to match coordinate {} to a coordinate ID", coordinate);
+      throw new NoDataException("No coordinate ID found for the given point: " + coordinate);
```
sebastian-peter: Logging an error is obsolete when we also throw and exception in the next line, imho

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -114,7 +115,18 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
+      throws NoDataException {
+
+    List<Point> invalidCoordinates =
+        coordinates.stream()
+            .filter(coordinate -> idCoordinateSource.getId(coordinate).isEmpty())
+            .toList();
+
+    if (!invalidCoordinates.isEmpty()) {
+      throw new NoDataException("No data for given coordinates: " + invalidCoordinates);
+    }
```
sebastian-peter: Hmm. I think there's a way to heal this instead of breaking right away.
I'd say, if for a coordinate we cannot retrieve weather data, we try to provide the last available data (the available data for the last point in time _before_ now) first. In this case, we also log a warning. 
If that also fails though, we should throw an exception.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -90,25 +91,63 @@ public Optional<Set<String>> getSourceFields() {
...
+      throws NoDataException {
+
+    List<Point> invalidCoordinates =
+        coordinates.stream()
+            .filter(coordinate -> !coordinateToTimeSeries.containsKey(coordinate))
+            .toList();
+
+    if (!invalidCoordinates.isEmpty()) {
+      throw new NoDataException("No data for given coordinates: " + invalidCoordinates);
+    }
```
sebastian-peter: Same as above

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -138,32 +151,52 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
       CompletableFuture<GetResult> futureResult =
           connector.get(generateWeatherKey(date, coordinateId.get()));
       GetResult getResult = futureResult.join();
       JsonObject jsonWeatherInput = getResult.contentAsObject();
-      return toTimeBasedWeatherValue(jsonWeatherInput);
+      return toTimeBasedWeatherValue(jsonWeatherInput)
+          .orElseThrow(
+              () ->
+                  new NoDataException(
+                      "No valid weather data found for the given date and coordinate."));
```
sebastian-peter, line 165: Same here: Maybe try to take an earlier data point instead?

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -90,25 +91,63 @@ public Optional<Set<String>> getSourceFields() {
...
+    if (timeSeries == null) {
+      throw new NoDataException("No weather data found for the given coordinate: " + coordinate);
+    }
+
+    return timeSeries
+        .getTimeBasedValue(date)
+        .orElseThrow(
+            () ->
+                new NoDataException(
+                    "No weather data found for the given coordinate: " + coordinate));
```
sebastian-peter: Same as above

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
sebastian-peter: It'd be great to have equivalent behavior across types of data sources, at least where it's possible

sebastian-peter (changes requested):
While looking through this, it came to my mind that some of these errors we could try and heal instead, before throwing an exception. 

I realize this requires a bit more effort, but maybe it still makes sense to handle it in this PR? On the other hand, SIMONA has to deal with these new types of exceptions anyway, and at a later step it would crash less because we healed the problem instead.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added weatherSource documentation [#1390](https://github.com/ie3-institute/PowerSystemDataModel/issues/1390)
 - Added standard asset parameter for `3wTransformer` in `ReadTheDocs` [#1417](https://github.com/ie3-institute/PowerSystemDataModel/issues/1417)
 - Added getter sRated for SystemParticipant inputs and updated them in tests in src[#1412](https://github.com/ie3-institute/PowerSystemDataModel/issues/1412)
+- Added explicit handling for cases where no weather data is received from any source [#554](https://github.com/ie3-institute/PowerSystemDataModel/issues/554)
```
danielfeismann: Please move this to the end of this list.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -27,12 +29,9 @@
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.tuple.Pair;
 import org.locationtech.jts.geom.Point;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /** Couchbase Source for weather data */
 public class CouchbaseWeatherSource extends WeatherSource {
-  private static final Logger logger = LoggerFactory.getLogger(CouchbaseWeatherSource.class);
   private static final String DEFAULT_TIMESTAMP_PATTERN = "yyyy-MM-dd'T'HH:mm:ssxxx";
```
danielfeismann: the timestamp pattern is also unused

src/main/java/edu/ie3/datamodel/io/source/WeatherSource.java
```diff
@@ -34,6 +35,58 @@ public abstract class WeatherSource extends EntitySource {
...
+    Duration step = Duration.between(stepReference, fallback);
+    if (step.isNegative()) {
+      // stepReference is after fallback — data inconsistency, accept as a safe default
+      log.warn(
+          "Unexpected step reference {} after fallback {} when checking fallback for {}. Accepting fallback unconditionally.",
+          stepReference,
+          fallback,
+          requested);
+      return true;
+    }
```
danielfeismann, line 78: If we encounter data inconsistencies, would it make sense to throw an exception here? It might not be the best place to do so, but if the times are mixed up like this, other things won't work as expected either.

src/main/java/edu/ie3/datamodel/io/source/WeatherSource.java
```diff
@@ -34,6 +35,58 @@ public abstract class WeatherSource extends EntitySource {
...
+   * determined (only one data point available before the requested time), the fallback is accepted.
+   *
+   * @param requested the originally requested time
+   * @param fallback the most recent known timestamp before the requested time
+   * @param stepReference the timestamp just before the fallback (used to infer step size), or
+   *     {@code null} if unavailable
+   * @return {@code true} if the fallback is within {@link #MAX_FALLBACK_STEPS} of the requested
+   *     time
+   */
+  protected static boolean isFallbackAcceptable(
```
danielfeismann, line 57: I would prefer to have some test on this method as well.

src/main/java/edu/ie3/datamodel/io/source/WeatherSource.java
```diff
@@ -34,6 +35,58 @@ public abstract class WeatherSource extends EntitySource {
...
+      // stepReference is after fallback — data inconsistency, accept as a safe default
+      log.warn(
+          "Unexpected step reference {} after fallback {} when checking fallback for {}. Accepting fallback unconditionally.",
+          stepReference,
+          fallback,
+          requested);
+      return true;
+    }
+    if (step.isZero()) {
+      // Two consecutive known timestamps are identical — step size cannot be determined
```
danielfeismann: I'm wondering if, in this case, we should use requested and fallback/stepReference to determine step size. What do you thing? Maybe I'm wrong.

danielfeismann: Ah, in that case step == gap and so we are always within the limits, right?

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvWeatherSourceIconTest.groovy
```diff
@@ -299,4 +300,112 @@ class CsvWeatherSourceIconTest extends Specification implements CsvTestDataMeta,
...
+
+    then:
+    def ex = thrown(NoDataException)
+    ex.message.contains("No weather data found for coordinate")
+    ex.message.contains("no earlier data available")
+  }
+
+  def "A CsvWeatherSource throws NoDataException when the fallback is beyond the maximum allowed steps"() {
+    given:
+    def farFutureTime = IconWeatherTestData.TIME_17H.plusHours(10)
```
danielfeismann: Just minor, however I guess it would be good to test closer to the edge of the max step size. E.g., 4 would also work, right? Same at the other tests.

src/test/groovy/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSourceCosmoIT.groovy
```diff
@@ -125,33 +126,84 @@ class InfluxDbWeatherSourceCosmoIT extends Specification implements TestContaine
...
 
-  def "An InfluxDbWeatherSource will return an equivalent to 'empty' when being unable to map a coordinate to its ID"() {
+  def "An InfluxDbWeatherSource will throw NoDataException when being unable to map a coordinate to its ID"() {
     given:
     def validCoordinate = CosmoWeatherTestData.COORDINATE_193186
     def invalidCoordinate = GeoUtils.buildPoint(7d, 48d)
     def time = CosmoWeatherTestData.TIME_15H
     def timeInterval = new ClosedInterval(CosmoWeatherTestData.TIME_15H, CosmoWeatherTestData.TIME_17H)
-    def emptyTimeSeries = new IndividualTimeSeries(UUID.randomUUID(), Collections.emptySet())
     def timeseries_193186 = new IndividualTimeSeries(null,
```
danielfeismann: this is no longer used.

src/test/groovy/edu/ie3/datamodel/io/source/sql/SqlWeatherSourceIconIT.groovy
```diff
@@ -109,6 +110,45 @@ class SqlWeatherSourceIconIT extends Specification implements TestContainerHelpe
...
+
+    when:
+    source.getWeather(farFutureTime, IconWeatherTestData.COORDINATE_67775)
+
+    then:
+    def ex = thrown(NoDataException)
+    ex.message.contains("No weather data found for coordinate")
+    ex.message.contains("exceeds the maximum fallback")
+  }
+
```
danielfeismann, line 152: Please add the case for returning partial results here as well. Or is there a reason that this is already covered?

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -97,25 +116,80 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
+      }
+
+      // Fallback: try the last known value before the requested date (within MAX_FALLBACK_STEPS)
+      String fallbackQuery = createQueryStringForLastBefore(date, coordinateId.get());
+      QueryResult fallbackResult = session.query(new Query(fallbackQuery));
+      List<TimeBasedValue<WeatherValue>> fallbackValues =
+          filterEmptyOptionals(optTimeBasedValueStream(fallbackResult))
+              .sorted((a, b) -> b.getTime().compareTo(a.getTime()))
+              .toList();
+      if (!fallbackValues.isEmpty()) {
```
danielfeismann: there are some duplications. May it is possible to reduce them? However, please check carefully if this makes sense or complicates this more than it helps...

danielfeismann (changes requested):
Thanks for including the steps for possible healing of missing data. I have some additions and questions.

src/main/java/edu/ie3/datamodel/io/source/WeatherSource.java
PhilippSchmelter: you are right. throwing here is cleaner.

src/main/java/edu/ie3/datamodel/io/source/WeatherSource.java
PhilippSchmelter: you are right! same result either way

src/test/groovy/edu/ie3/datamodel/io/source/csv/CsvWeatherSourceIconTest.groovy
PhilippSchmelter: also true. i changed to 3 and 4

src/test/groovy/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSourceCosmoIT.groovy
PhilippSchmelter: removed

src/test/groovy/edu/ie3/datamodel/io/source/sql/SqlWeatherSourceIconIT.groovy
PhilippSchmelter: Good Point!. Ive overlooked that. Added it to both Influx Tests too

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
PhilippSchmelter: i extracted it into a little helper method just for InfluxDb. Implementing that helper across all Sources wouldnt be clean in my opinion. Adding that Method to the BaseClass, would just thicken and couple the classes unnecessary

danielfeismann (commented):
Thanks for considering the discussed points. This looks good to me! However, since weather is very important within our simulations I would prefer to have also a look by @staudtMarius on this.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -79,31 +80,121 @@ public void validate() throws ValidationException {
...
     return coordinateToTimeSeries.entrySet().stream()
-        .collect(Collectors.toMap(Map.Entry::getKey, t -> t.getValue().getTimeKeysAfter(time)));
+        .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getTimeKeysAfter(time)))
+        .entrySet()
+        .stream()
+        .filter(e -> !e.getValue().isEmpty())
+        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
+
+  @Override
```
staudtMarius, line 153: Why did you change the order of the methods?

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -107,16 +105,31 @@ public void validate() throws ValidationException {
...
+    if (coordinates.isEmpty())
+      throw new NoDataException("No coordinates provided for weather data query.");
+
+    List<Point> unknownCoordinates =
+        coordinates.stream()
+            .filter(coordinate -> idCoordinateSource.getId(coordinate).isEmpty())
+            .toList();
+
+    if (!unknownCoordinates.isEmpty())
+      log.warn(
```
staudtMarius, line 117: You are not excluding these coordinates from checking.

src/main/java/edu/ie3/datamodel/io/source/influxdb/InfluxDbWeatherSource.java
```diff
@@ -82,10 +86,24 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
         coordinates.stream().collect(Collectors.toMap(point -> point, idCoordinateSource::getId));
+
+    List<Point> unknownCoordinates =
+        coordinatesToId.entrySet().stream()
+            .filter(entry -> entry.getValue().isEmpty())
+            .map(Map.Entry::getKey)
+            .toList();
+
+    if (!unknownCoordinates.isEmpty())
+      log.warn(
```
staudtMarius, line 103: You are not skipping these coordinates with you current implementation.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -103,22 +109,36 @@ public Map<Point, IndividualTimeSeries<WeatherValue>> getWeather(
...
+    if (coordinates.isEmpty())
+      throw new NoDataException("No coordinates provided for weather data query.");
+
+    List<Point> unknownCoordinates =
+        coordinates.stream()
+            .filter(coordinate -> idCoordinateSource.getId(coordinate).isEmpty())
+            .toList();
+
+    if (!unknownCoordinates.isEmpty())
+      log.warn(
```
staudtMarius, line 136: Same as in the other database sources.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -155,10 +185,52 @@ public Optional<TimeBasedValue<WeatherValue>> getWeather(ZonedDateTime date, Poi
                   ps.setTimestamp(2, Timestamp.from(date.toInstant()));
                 }));
 
-    if (timeBasedValues.isEmpty()) return Optional.empty();
-    if (timeBasedValues.size() > 1)
-      log.warn("Retrieved more than one result value, using the first");
-    return Optional.of(timeBasedValues.getFirst());
+    if (!timeBasedValues.isEmpty()) {
```
staudtMarius, line 177: This is the same as in the other sources. Can you abstract these lines and put them into the abstract class `WeatherSource`?

staudtMarius (changes requested):
Here are some first remarks from my side.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
PhilippSchmelter: could you please specify that question? I checked the diff against dev and only changes i found are that the methods we used to have are a bit longer here and there. Also we got a new getTimeKeysAfter Method, but no real change in order. Please correct me if im wrong

staudtMarius: Ok. I mistook this for another method based on the input parameters.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Create data model for air condition [#1396](https://github.com/ie3-institute/PowerSystemDataModel/issues/1396)
 - Added attribute units for standard lines types [#1483](https://github.com/ie3-institute/PowerSystemDataModel/issues/1483)
 - Add validation methods to `DataSource` [#1559](https://github.com/ie3-institute/PowerSystemDataModel/issues/1559)
+- Added explicit handling for cases where no weather data is received from any source [#554](https://github.com/ie3-institute/PowerSystemDataModel/issues/554)
```
staudtMarius: Could you move this to the end?

src/main/java/edu/ie3/datamodel/exceptions/NoDataException.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * © 2024. TU Dortmund University,
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.exceptions;
+
+/**
+ * Exception that should be used whenever no data is received{@link
```
staudtMarius: I think you are missing a word here.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -95,66 +93,179 @@ public void validate() throws ValidationException {
...
+   * WeatherSource#MAX_FALLBACK_STEPS} time steps of the requested date.
+   *
+   * @param date the requested date
+   * @param coordinate the coordinate point
+   * @param coordinateId the coordinate ID
+   * @return the last known weather value before date
+   * @throws NoDataException if no earlier data is available or the gap exceeds the allowed steps
+   * @throws SourceException if the stored data points are inconsistently ordered
+   */
+  private TimeBasedValue<WeatherValue> getLastKnownWeather(
```
staudtMarius, line 184: SInce the checking of the fallback values is simmilar for all the weather source implementation, I think we schould abstract them and move this abtract method to the base class.
But this could also be done in another issue.

staudtMarius (changes requested):
Some remarks from my side.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
PhilippSchmelter: good point! this saves us a lot of lines :D

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
```diff
@@ -79,31 +80,83 @@ public void validate() throws ValidationException {
...
+    Optional<TimeBasedValue<WeatherValue>> exact = timeSeries.getTimeBasedValue(date);
+    if (exact.isPresent()) {
+      return exact.get();
+    }
+
+    List<TimeBasedValue<WeatherValue>> fallbacks = new ArrayList<>(2);
+    timeSeries
+        .getPreviousDateTime(date)
+        .flatMap(timeSeries::getTimeBasedValue)
+        .ifPresent(fallbacks::add);
```
staudtMarius: I think that we will add more than just two elements here. Have you checked, if this could lead to a problem?

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
```diff
@@ -155,10 +174,24 @@ public Optional<TimeBasedValue<WeatherValue>> getWeather(ZonedDateTime date, Poi
...
```
staudtMarius, line 177: The first if statement can be removed.

staudtMarius (commented):
I have two small remarks.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
PhilippSchmelter: The if (!timeBasedValues.isEmpty()) is needed. Right below it is return timeBasedValues.get(0), which would throw IOOB on an empty list. Also, the fallback query and applyFallbackOrThrow only run because of this if. Without it the early return always fires and the fallback path is dead code.

Please correct me if i am wrong here! :)

src/main/java/edu/ie3/datamodel/io/source/csv/CsvWeatherSource.java
PhilippSchmelter: Actually it was capped at 2, but fair point it wasn't really clear. Refactored it to a for loop with i < 2 so the cap is obvious now.

src/main/java/edu/ie3/datamodel/io/source/sql/SqlWeatherSource.java
staudtMarius: I checked it again. Your right, we need the if statement.

staudtMarius (approved):

PR #1188, opened 2024-11-27
Ensure that thermal boundaries of thermal house are not the same

### LINKED ISSUE #1186: Ensure that thermal boundaries of thermal house are not the same
'upperTemperatureLimit', 'targetTemperature' and 'lowerTemperatureLimit' should not be equal. 

This should be checked within PSDM

Edit: As well for ThermalStorage with equal inlet and return temperature. Even this is working, the storage make no sense in that case...

danielfeismann:
resolves #1186


sebastian-peter (approved):
LGTM

PR #1193, opened 2024-11-29
Renamed timeseries mapping `participant` column to `asset`.

### LINKED ISSUE #1087: Check if `MappingEntry` should still be a record
With #1085, we added getters to `MappingEntry` to fix writing them out. Examine if it still makes sense to have them as `records` or if a plain class would fit better, or if there is another solution.

### LINKED ISSUE #1191: Change timeseries mapping column naming
After #1130 is merged, we can provide values to other assets than participants. therefore the naming of the column should be changed.

staudtMarius:
Resolves #1087
Resolves #1191


sebastian-peter (approved):
LGTM

PR #1231, opened 2025-02-03
Ms/#1230 add bdew load profile time series

staudtMarius:
Resolves #1230

PR #1233, opened 2025-02-03
Added `RandomLoadProfileTimeSeries`

### LINKED ISSUE #1232: Add RandomLoadProfileTimeSeries

staudtMarius:
Resolves #1232

This branch is build on top of #1231.


staudtMarius:
I think we can ignore the `sonarqube` issues regarding the random number generator, since we do not use the generated number in a security-sensitive case (see [here](https://stackoverflow.com/questions/62036514/make-sure-that-using-this-pseudorandom-number-generator-is-safe-here)).

danielfeismann:
Yes, but please have a look if you could reduce the number of duplicated lines.

src/main/java/edu/ie3/datamodel/io/factory/timeseries/RandomLoadProfileFactory.java
```diff
@@ -0,0 +1,101 @@
...
+  public ComparableQuantity<Power> calculateMaxPower(
+      RandomLoadProfile loadProfile, Set<LoadProfileEntry<RandomLoadValues>> loadProfileEntries) {
+    return Quantities.getQuantity(159d, WATT);
+  }
+
+  @Override
+  public ComparableQuantity<Energy> getLoadProfileEnergyScaling(RandomLoadProfile loadProfile) {
+    return Quantities.getQuantity(716.5416966513656, PowerSystemUnits.KILOWATTHOUR);
+  }
+}
```
danielfeismann, line 101: Why are we returning hard coded values here?

src/main/java/edu/ie3/datamodel/models/timeseries/repetitive/LoadProfileTimeSeries.java
```diff
@@ -46,8 +48,13 @@ public LoadProfileTimeSeries(
             .collect(
                 Collectors.toMap(LoadProfileEntry::getQuarterHour, LoadProfileEntry::getValue));
 
-    this.maxPower = maxPower;
-    this.profileEnergyScaling = profileEnergyScaling;
+    // use default value is null value is given
```
danielfeismann, line 51:
```suggestion
    // use default value if null value is given
```

just a typo?

src/main/java/edu/ie3/datamodel/models/timeseries/repetitive/LoadProfileTimeSeries.java
```diff
@@ -46,8 +48,13 @@ public LoadProfileTimeSeries(
...
+    this.maxPower =
+        Optional.ofNullable(maxPower)
+            .orElseGet(() -> Quantities.getQuantity(0d, PowerSystemUnits.KILOWATT));
```
danielfeismann, line 54: Is this a proper default value?

src/main/java/edu/ie3/datamodel/models/value/load/RandomLoadValues.java
```diff
@@ -1,78 +1,81 @@
 /*
- * © 2021. TU Dortmund University,
+ * © 2024. TU Dortmund University,
```
danielfeismann, line 2: I would keep it as it is, but very minor thing

danielfeismann (changes requested):
Some thoughts from my side. 

Another thing: Should we / can we check the profiles (random, but also the others we included) with SystemParticipantValidationUtils.checkLoad() ?

src/main/java/edu/ie3/datamodel/models/timeseries/repetitive/LoadProfileTimeSeries.java
staudtMarius: I will change this.

staudtMarius:
> > @danielfeismann We do check the load profile in `SystemParticipantValidationUtils.checkLoad()`:
> 
> Yes, I know. I just wanted to ask if we apply this validation as well to the loadProfiles within your tests.

What do you mean with `within your tests`?

danielfeismann (approved):
LGTM

PR #1252, opened 2025-02-21
Introduce thermal power for thermal storages

### LINKED ISSUE #679: Charge / discharge power for thermal storages
Currently only capacity limits are defined for thermal storages, however, for practical usage also the rate of change of stored capacity aka. charge / discharge power are of relevance. Please add an attribute for it.

danielfeismann:
resolves #679


staudtMarius (changes requested):
Just two small remoarks.

staudtMarius (approved):
LGTM

PR #1269, opened 2025-02-28
Copy builders return this instance

### LINKED ISSUE #1250: Refactor `CopyBuilder`s to use `thisInstance` to return itself.
This is the behaviour we want:
Image

but there are many cases, where we have the following code:
Image

Therefore, all part should be adapted to be like the first example.

marvinheintze:
Resolves #1250 
 Replaced return this with return `thisInstance()` in CopyBuilders


staudtMarius:
@marvinheintze The following copy builders needs to be adapted as well:
- LoadInput
- MeasurementUnitInput
- NodeGraphicInput
- NodeInput
- OperatorInput
- PvInput
- StorageTypeInput
- SwitchInput
- ThermalHouseInput
- Transformer2WInput
- Transformer3WInput
- WecInput
- WecTypeInput

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaced `LoadProfileInput` with `LoadProfileTimeSeries` [#1228](https://github.com/ie3-institute/PowerSystemDataModel/issues/1228)
 - Enhance `CsvDataSource` [#1246](https://github.com/ie3-institute/PowerSystemDataModel/issues/1246)
 - Updated `_joint_grid` csv files from simona [#750](https://github.com/ie3-institute/PowerSystemDataModel/issues/750)
+- Replaced `return this` with `return thisInstance` in CopyBuilders [#1250](https://github.com/ie3-institute/PowerSystemDataModel/issues/1250)
```
staudtMarius, line 46: Can you move this to the SNAPSHOT section?

src/main/java/edu/ie3/datamodel/models/input/container/InputContainer.java
```diff
@@ -30,5 +30,7 @@ interface InputContainerCopyBuilder<T extends UniqueInputEntity> {
 
     /** Returns the altered {@link InputContainer} */
     InputContainer<T> build() throws ValidationException;
+
+    InputContainerCopyBuilder<T> thisInstance();
```
staudtMarius, line 34: We could change this interface to be an abstract class. If we do, we could make the method `thisInstance()` protected.

staudtMarius (changes requested):
Looks good, I just have two small requests.

src/main/java/edu/ie3/datamodel/models/input/container/JointGridContainer.java
```diff
@@ -107,7 +107,7 @@ protected JointGridContainerCopyBuilder(JointGridContainer jointGridContainer) {
     }
 
     @Override
-    protected JointGridContainerCopyBuilder thisInstance() {
+    public JointGridContainerCopyBuilder thisInstance() {
       return this;
     }
```
staudtMarius, line 112: Can you make this method protected as well?

src/main/java/edu/ie3/datamodel/models/input/container/SubGridContainer.java
```diff
@@ -105,7 +105,7 @@ public SubGridContainerCopyBuilder subnet(int subnet) {
     }
 
     @Override
-    protected SubGridContainerCopyBuilder thisInstance() {
+    public SubGridContainerCopyBuilder thisInstance() {
       return this;
     }
```
staudtMarius, line 110: Can you make this method protected again?

src/main/java/edu/ie3/datamodel/models/input/container/SystemParticipants.java
```diff
@@ -404,6 +404,11 @@ public SystemParticipantsCopyBuilder storages(Set<StorageInput> storages) {
      */
     public SystemParticipantsCopyBuilder wecPlants(Set<WecInput> wecPlants) {
       this.wecPlants = wecPlants;
+      return (SystemParticipantsCopyBuilder) thisInstance();
+    }
+
+    @Override
+    protected InputContainerCopyBuilder<SystemParticipantInput> thisInstance() {
```
staudtMarius, line 411: You should set the returned type to `SystemParticiapntCopyBuilder` and them remove the type casts.

staudtMarius (changes requested):
There are some more methods in `SystemParticipantCopyBuilder`, that are not using the `thisInstance` method.

PR #1281, opened 2025-03-11
Enhancing value retrieval in `TimeSeriesSource`.

### LINKED ISSUE #1280: Enhance value retrieval in `TimeSeriesSource`
Thanks for putting in the effort. I've been looking through the code and kept thinking that we should just integrate this into `getTimeKeysAfter` and make it the PSDM's job.... That'd make it less complicated here.

_Originally posted by @sebastian-peter in https://github.com/ie3-institute/simona/pull/1038#pullrequestreview-2674212572_

staudtMarius:
Resolves #1280

### DIFF
--- a/src/main/java/edu/ie3/datamodel/io/source/sql/SqlTimeSeriesSource.java
+++ b/src/main/java/edu/ie3/datamodel/io/source/sql/SqlTimeSeriesSource.java
@@ -199,6 +199,15 @@ public class SqlTimeSeriesSource<V extends Value> extends TimeSeriesSource<V> {
         .toList();
   }
 
+  @Override
+  public Optional<ZonedDateTime> getLastTimeKeyBefore(ZonedDateTime time) {
+    return dataSource
+        .executeQuery(
+            queryForValueBefore, ps -> ps.setTimestamp(1, Timestamp.from(time.toInstant())))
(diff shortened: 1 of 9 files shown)


src/main/java/edu/ie3/datamodel/io/source/sql/SqlTimeSeriesSource.java
```diff
@@ -199,6 +199,15 @@ public List<ZonedDateTime> getTimeKeysAfter(ZonedDateTime time) {
         .toList();
   }
 
+  @Override
+  public Optional<ZonedDateTime> getLastTimeKeyBefore(ZonedDateTime time) {
+    return dataSource
+        .executeQuery(
+            queryForValueBefore, ps -> ps.setTimestamp(1, Timestamp.from(time.toInstant())))
+        .map(valueFactory::extractTime)
+        .findFirst();
```
danielfeismann, line 208: is the stream ordered reverse or why does .findFirst() results in the last timeKey before?

staudtMarius: I improved the code, to address this point.

danielfeismann (approved):
LGTM

PR #1295, opened 2025-04-04
Enhancing load profile source.

### LINKED ISSUE #1294: Enhance load profile sources
The enhancement should consider the following points:
- return a sample function for `RandomLoadProfile` to sample an own `PValue` for each `RandomLoadModel`
- consider last value (can be used by MarkowLoadProfile in the future)

staudtMarius:
Resolves #1294


danielfeismann (changes requested):

staudtMarius:
This PR can be closed, since it will be unnecessary after merging #1439.

PR #1297, opened 2025-04-07
TimeSeriesMappingSource getMapping() error handling + tests

### LINKED ISSUE #1287: Invalid time series mapping values just disappear
If there's an error during mapping entry creation, the failure is just filtered out. It should be thrown instead.

https://github.com/ie3-institute/PowerSystemDataModel/blob/6b3b06b287cad9bbb095e7f02c877d3ffddf760f/src/main/java/edu/ie3/datamodel/io/source/TimeSeriesMappingSource.java#L43-L50

marvinheintze:
Resolves #1287 
Added error handling for getMapping() and added tests for the error handling

### DIFF
--- a/src/test/groovy/edu/ie3/datamodel/io/source/TimeSeriesMappingSourceTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/io/source/TimeSeriesMappingSourceTest.groovy
@@ -5,11 +5,14 @@
  */
 package edu.ie3.datamodel.io.source
 
+import edu.ie3.datamodel.exceptions.FactoryException
+import edu.ie3.datamodel.exceptions.FailureException
 import edu.ie3.datamodel.exceptions.SourceException
 import spock.lang.Shared
 import spock.lang.Specification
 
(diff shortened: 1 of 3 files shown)


src/test/groovy/edu/ie3/datamodel/io/source/TimeSeriesMappingSourceTest.groovy
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  def "should throw SourceException for invalid asset"(){
+    given: "DummyMappingSource"
+    validSource = new MixedMappingSource(false, true)
+    when:
+    validSource.getMapping()
+
+    then:
+    thrown(SourceException)
```
staudtMarius, line 96: Could you check the message of the exception here and at the other places

staudtMarius (changes requested):
One small question from my side.

staudtMarius (approved):
LGTM. I just improved the checking for errors a bit.

PR #1332, opened 2025-05-22
Fixed em fields in input models.

### LINKED ISSUE #1331: Input and output field for ems in ParticipantInput and EmInput do not match
We expect the em field for participant inputs to be `em`. For em input, we expect the field `parent_em`. When persisting these models with a sink. These fields are persisted a `controlling_em`.

If we persist one of these models and read them again, the em information is dropped.

staudtMarius:
Resolves #1331


danielfeismann (approved):
LGTM

PR #1358, opened 2025-07-03
Enhance ValidationUtils for load profile

### LINKED ISSUE #1357: Enhance ValidationUtils for load profile

danielfeismann:
resolves #1357 

merge first
- #1374


staudtMarius:
@danielfeismann I see one big problem with this validation. If we want to support custom load profiles, we have to either remove this validation later or adpat it, so that no exception is thrown.

danielfeismann:
Which custom load profiles do you think of?

staudtMarius:
Custom might not be the best name for it. I though, that someone might want to provide and use a load profile time series, that is not linked to a load profile, that is defined in the PSDM.

danielfeismann:
Ok, my understanding of the [docs](https://powersystemdatamodel.readthedocs.io/en/latest/models/input/participant/load.html#load-lp) is that one should use LoadProfile#NO_LOAD_PROFILE for this purpose.

staudtMarius:
If you use for example primary data, you can add LoadProfile#NO_LOAD_PROFILE. But if we want to use the load profile service, we need to specify a profile in the load input.

staudtMarius (approved):
LGTM.

PR #1396, opened 2025-07-25
Create data model for air condition

### LINKED ISSUE #1352: Create data model for Air Condition
The heat pump model can be used as orientation and basis

- AirConditionInput
- AirConditionTypeInput
- AirConditoinResult

danielfeismann:
resolves #1352


src/main/java/edu/ie3/datamodel/io/factory/input/participant/AcInputFactory.java
```diff
@@ -0,0 +1,48 @@
...
+    return new AcInput(
+        uuid,
+        id,
+        operator,
+        operationTime,
+        node,
+        thermalBusInput,
+        qCharacteristics,
+        em,
+        (AcTypeInput) typeInput);
```
staudtMarius: Is there a better way than type casting?

staudtMarius (changes requested):
Looks good to me. I have just one small question.

src/main/java/edu/ie3/datamodel/io/factory/input/participant/AcInputFactory.java
danielfeismann: good point.

staudtMarius (approved):
LGTM.

PR #1410, opened 2025-08-04
Change spotless to use googleJavaFormat('1.28.0')

danielfeismann:
resolves #1409

PR #1411, opened 2025-08-04
Update to Java21

### LINKED ISSUE #1408: Check Java Version 21

danielfeismann:
resolves #1408 

merge first
- #1410


staudtMarius:
@danielfeismann Could you adapt the `getting started` page in the docs. The new requirements should be `Java >= 21`.

docs/readthedocs/gettingstarted.md
```diff
@@ -5,7 +5,7 @@ If you feel, something is missing, please contact us!
 
 ## Requirements
 
-Java > v 17
+Java > v 21
```
staudtMarius: Should we write `>= v 21`? There is no need to use a java version greater than 21, if you want to add the PSDM as a dependency.

src/main/java/edu/ie3/datamodel/io/factory/timeseries/SqlIdCoordinateFactory.java
```diff
@@ -20,13 +20,13 @@
...
   private static final String COORDINATE_ID = "id";
   private static final String COORDINATE = "coordinate";
-  private final WKBReader reader = new WKBReader();
 
   @Override
   protected IdCoordinateInput buildModel(SimpleFactoryData data) {
     try {
       int coordinateId = data.getInt(COORDINATE_ID);
       byte[] byteArr = WKBReader.hexToBytes(data.getField(COORDINATE));
+      WKBReader reader = new WKBReader();
```
staudtMarius, line 29: Why did you change this?

staudtMarius (changes requested):
Two small questions from my side.

src/main/java/edu/ie3/datamodel/io/factory/timeseries/SqlIdCoordinateFactory.java
danielfeismann: WKBReader is not thread-safe. This fixes that, by always using a new reader. Keeping it will let some of the test fail in Java21 (at least it failed when i was testing it first)

staudtMarius (approved):
LGTM.

PR #1415, opened 2025-08-05
Fix transformer susceptance in readTheDocs to negative values

### LINKED ISSUE #1078: Fix transformer susceptance in readTheDocs to negative values
with #378 we expect transformer susceptance to be negative. Thus the documentation should be adapted accordingly.

danielfeismann:
resolves #1078


staudtMarius (approved):
LGTM

PR #1429, opened 2025-09-01
Correctly persist EmInput entities in CsvFileSink

### LINKED ISSUE #1337: EmInputs aren't written while persisting a `JointGridContainer`
When calling the method `persistJointGrid` of `CsvFileSink`, no `EmInputs` are persisted. This leads to missing ems after reading the grid.

pierrepetersmeier:
Resolves #1337


staudtMarius:
@pierrepetersmeier Could you add a test that reads in the persisted grid and compares it with the original grid? There are some test for grids without ems in `GridIoIT`. Maybe you could simply add some ems to this existing grid.

danielfeismann:
Is this complete and only waiting for some review? @pierrepetersmeier

pierrepetersmeier:
complete :)

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -225,14 +226,29 @@ public void persistJointGrid(JointGridContainer jointGridContainer) {
             .flatMap(Optional::stream)
             .collect(Collectors.toSet());
 
+    // extract all EmInput entities
+    Set<EmInput> emInputs =
+        systemParticipants.allEntitiesAsList().stream()
+            .filter(Objects::nonNull)
+            .map(HasEm.class::cast)
```
danielfeismann: if you filter for the HasEm.class before, (HasEm.class::isInstance)
you only cast, if its type-safe.

danielfeismann (changes requested):
Looks good except only little detail that may can be improved.

danielfeismann:
> @pierrepetersmeier Could you add a test that reads in the persisted grid and compares it with the original grid? There are some test for grids without ems in `GridIoIT`. Maybe you could simply add some ems to this existing grid.

@pierrepetersmeier: is that done? Doesn't look like, to be honest :)

src/test/groovy/edu/ie3/datamodel/io/sink/CsvFileSinkTest.groovy
```diff
@@ -15,6 +15,7 @@ import edu.ie3.datamodel.io.processor.input.InputEntityProcessor
 import edu.ie3.datamodel.io.processor.result.ResultEntityProcessor
 import edu.ie3.datamodel.io.processor.timeseries.TimeSeriesProcessor
 import edu.ie3.datamodel.io.processor.timeseries.TimeSeriesProcessorKey
+import edu.ie3.datamodel.io.source.csv.CsvJointGridContainerSource
```
danielfeismann: This import is not used.

src/test/groovy/edu/ie3/datamodel/io/source/csv/GridIoIT.groovy
```diff
@@ -41,7 +40,7 @@ class GridIoIT extends Specification implements CsvTestDataMeta {
     sinkHierarchic = new CsvFileSink(tempDirectory.toAbsolutePath(), hierarchicNamingStrategy, ",")
   }
 
-  def cleanupSpec() {
```
danielfeismann, line 44: What are the reason behind these changes?

src/test/groovy/edu/ie3/datamodel/io/source/csv/GridIoIT.groovy
```diff
@@ -93,4 +93,53 @@ class GridIoIT extends Specification implements CsvTestDataMeta {
     then:
     thrown(FileException)
   }
+
+  def "Input flat JointGridContainer with EmInput equals Output flat JointGridContainer."() {
```
danielfeismann: It was really good that you added this test, since it disclosed that this is already covered by the first test in this file. So please remove this one and find out why persisting is not working yet (first test fails).

danielfeismann (changes requested):
I added the legacy energyManagementUnits which somehow got lost in another PR but should be here. However now some test is failing. It would be great if you could have a look on that.

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -228,14 +229,34 @@ public void persistJointGrid(JointGridContainer jointGridContainer) {
...
+        systemParticipants.allEntitiesAsList().stream()
+            .filter(HasEm.class::isInstance)
+            .map(HasEm.class::cast)
+            .map(HasEm::getControllingEm)
+            .flatMap(Optional::stream)
+            .flatMap(
+                em ->
+                    Stream.iterate(
+                        em, Objects::nonNull, current -> current.getControllingEm().orElse(null)))
+            .collect(Collectors.toSet());
```
danielfeismann: Ah, this is where it got wrong. Since the EMs are no longer systemParticipants this will always be empty. And that caused that the EMs did not get persisted. So this can be removed.

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -228,14 +229,34 @@ public void persistJointGrid(JointGridContainer jointGridContainer) {
...
+
+    Set<EmInput> containerEmInputs = new HashSet<>(jointGridContainer.getEmUnits().getEmUnits());
```
danielfeismann: Thats the line that was missing! Good solution. However, I prefer to keep the former naming of `emInputs`.

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -228,14 +229,34 @@ public void persistJointGrid(JointGridContainer jointGridContainer) {
...
+
+    Set<EmInput> allEmInputs = new HashSet<>(containerEmInputs);
+    allEmInputs.addAll(emInputs);
```
danielfeismann: Since the 'emInputs' from above are always empty this can be removed.

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
```diff
@@ -228,14 +229,34 @@ public void persistJointGrid(JointGridContainer jointGridContainer) {
...
     Stream.of(
             rawGridElements.allEntitiesAsList(),
             systemParticipants.allEntitiesAsList(),
             graphicElements.allEntitiesAsList(),
             types,
-            operators)
+            operators,
+            allEmInputs)
         .flatMap(Collection::stream)
+        .collect(Collectors.toSet())
```
danielfeismann: Not needed in my point of view

danielfeismann (changes requested):
Great, you found it! Just minor improvements on CsvFileSink from my side. Please also have a look on the sonarcube issue and have a look into my previous comments above.

src/main/java/edu/ie3/datamodel/io/sink/CsvFileSink.java
danielfeismann: Still not necessary in my point of view :). If you use `jointGridContainer.getEmUnits().getEmUnits().stream().toList(),` four lines above you can rollback all other changes in this file.

src/main/java/edu/ie3/datamodel/models/input/container/EnergyManagementUnits.java
```diff
@@ -0,0 +1,103 @@
...
+
+import edu.ie3.datamodel.models.input.EmInput;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/** Represents the accumulation of energy management units */
+public class EnergyManagementUnits implements InputContainer<EmInput> {
+
+  protected final Map<UUID, EmInput> emUnits;
```
danielfeismann: check the sonoarqube issue here. It was not caused by you however it would be great to fix this.

src/test/groovy/edu/ie3/datamodel/io/source/csv/GridIoIT.groovy
```diff
@@ -23,16 +28,11 @@ import java.nio.file.Path
  */
 class GridIoIT extends Specification implements CsvTestDataMeta {
 
-  @Shared
```
danielfeismann, line 26: what caused these changes?

src/test/groovy/edu/ie3/datamodel/io/source/csv/GridIoIT.groovy
```diff
@@ -10,6 +10,11 @@ import edu.ie3.datamodel.io.naming.DefaultDirectoryHierarchy
 import edu.ie3.datamodel.io.naming.EntityPersistenceNamingStrategy
 import edu.ie3.datamodel.io.naming.FileNamingStrategy
 import edu.ie3.datamodel.io.sink.CsvFileSink
+import edu.ie3.datamodel.models.OperationTime
+import edu.ie3.datamodel.models.input.EmInput
+import edu.ie3.datamodel.models.input.OperatorInput
+import edu.ie3.datamodel.models.input.container.JointGridContainer
+import edu.ie3.datamodel.models.input.container.SystemParticipants
```
danielfeismann: Why are these know necessary here?

danielfeismann (changes requested):
Almost there. I pushed a little addition to the docs and did minor adaptions. Besides that, there are only little things which would be nice to get included.

src/test/groovy/edu/ie3/datamodel/io/source/csv/GridIoIT.groovy
danielfeismann: also below

pierrepetersmeier: For the test isolation.

danielfeismann (approved):
LGTM!

PR #1439, opened 2025-09-17
Added abstraction for power value sources.

### LINKED ISSUE #1438: Abstracting power value sources
Since we are planing to introduce further sources for power values that does not fit into the current sources, an abstract source should be implemented that contains basic functionality.

staudtMarius:
Resolves #1438

Merge #1450 first.


src/main/java/edu/ie3/datamodel/io/source/PowerValueSource.java
```diff
@@ -0,0 +1,91 @@
...
+  non-sealed interface MarkovBased extends PowerValueSource<PowerProfile, InputData> {}
+
+  // -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+  // input data
+
+  /**
+   * Interface for the input data of {@link #getValue(InputData)}. The data is used to determine the
+   * next power.
+   */
+  sealed interface InputData permits PowerValueSource.TimeSeriesInputValue {
```
sebastian-peter: Maybe we can think about a different name for the interface here. `InputData` seems quite generic. Spontaneously I'd think of something that includes `Identifier`, maybe `PowerValueIdentifier`, or something better?

src/main/java/edu/ie3/datamodel/io/source/PowerValueSource.java
```diff
@@ -0,0 +1,91 @@
...
+import java.time.ZonedDateTime;
+import java.util.Optional;
+import java.util.function.Supplier;
+import javax.measure.quantity.Energy;
+import javax.measure.quantity.Power;
+import tech.units.indriya.ComparableQuantity;
+
+/** Interface defining base functionality for power value sources. */
+public sealed interface PowerValueSource<
+        P extends PowerProfile, ID extends PowerValueSource.InputData>
```
sebastian-peter: Please check out the sonarqube issue

src/main/java/edu/ie3/datamodel/io/naming/timeseries/LoadProfileMetaInformation.java
```diff
@@ -37,18 +32,11 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(super.hashCode(), profile);
+    return Objects.hash(profile);
```
sebastian-peter: Wouldn't do this, especially because the equals method still relies on the super class

src/main/java/edu/ie3/datamodel/io/source/PowerValueSource.java
```diff
@@ -0,0 +1,91 @@
...
+  /** Returns the profile of this source. */
+  P getProfile();
+
+  /**
+   * Method to get the next power value based on the provided input data.
+   *
+   * @param data input data that is used to calculate the next power value.
+   * @return an option for the power value.
+   */
+  Optional<PValue> getValue(ID data);
```
sebastian-peter: I think this method now becomes obsolete with the addition of `getValueSupplier`. We can just use a supplier all the time, even if it always returns the same value.

src/main/java/edu/ie3/datamodel/io/source/csv/CsvLoadProfileSource.java
```diff
@@ -61,24 +61,28 @@ public void validate() throws ValidationException {
...
-  public P getLoadProfile() {
-    return getTimeSeries().getLoadProfile();
+  public Supplier<Optional<PValue>> getValueSupplier(TimeSeriesInputValue data) {
+    return loadProfileTimeSeries.supplyValue(data.time());
+  }
+
+  @Override
+  public P getProfile() {
+    return loadProfileTimeSeries.getLoadProfile();
   }
```
sebastian-peter: Unnecessary ovveride imho

src/main/java/edu/ie3/datamodel/io/source/sql/SqlLoadProfileSource.java
```diff
@@ -101,39 +95,39 @@ public void validate() throws ValidationException {
...
+  public Supplier<Optional<PValue>> getValueSupplier(TimeSeriesInputValue data) {
+    ZonedDateTime time = data.time();
+    Optional<LoadValues<P>> loadValueOption = queryForValue(time);
+
+    if (loadValueOption.isPresent()) {
+      LoadValues<P> loadValue = loadValueOption.get();
+      return () -> Optional.of(loadValue.getValue(time, profile));
+    } else {
+      return Optional::empty;
+    }
```
sebastian-peter: Maybe you can use the map function on the optional here?

src/main/java/edu/ie3/datamodel/models/timeseries/TimeSeries.java
```diff
@@ -77,15 +77,7 @@ public Optional<TimeBasedValue<R>> getTimeBasedValue(ZonedDateTime time) {
...
    */
-  protected abstract Optional<ZonedDateTime> getNextDateTime(ZonedDateTime time);
-
-  /**
-   * Get all {@link ZonedDateTime}s after the given time.
-   *
-   * @param time given time
-   * @return a list of all time keys
-   */
-  public abstract List<ZonedDateTime> getTimeKeysAfter(ZonedDateTime time);
```
sebastian-peter, line 88: Yes, finally 👍

src/main/java/edu/ie3/datamodel/models/timeseries/individual/IndividualTimeSeries.java
```diff
@@ -60,13 +60,18 @@ public Optional<ZonedDateTime> getPreviousDateTime(ZonedDateTime time) {
...
-  @Override
+  /**
+   * Get all {@link ZonedDateTime}s after the given time.
+   *
+   * @param time given time
+   * @return a list of all time keys
+   */
   public List<ZonedDateTime> getTimeKeysAfter(ZonedDateTime time) {
     return timeToValue.keySet().stream().filter(timeKey -> timeKey.isAfter(time)).sorted().toList();
   }
```
sebastian-peter, line 77: Should me mark this method as obsolete, or just remove it completely? Or are there reasons to keep it?

sebastian-peter (changes requested):
Some comments and questions from my side

src/main/java/edu/ie3/datamodel/models/timeseries/TimeSeries.java
staudtMarius: What do you mean with `finally`? I think we added this method not long ago. I just moved this method to another place.

src/main/java/edu/ie3/datamodel/models/timeseries/individual/IndividualTimeSeries.java
staudtMarius: I think that we are using this method in SIMONA. But I will check this later and remove it, if we don't use it.

src/main/java/edu/ie3/datamodel/models/timeseries/TimeSeries.java
sebastian-peter: Thought we would get rid of this method...

src/main/java/edu/ie3/datamodel/models/timeseries/individual/IndividualTimeSeries.java
sebastian-peter: This would be replaced by `getNextDateTime` in SIMONA I think. But we don't have to delete this method right away.

staudtMarius: This method is actually used in the `PrimaryServiceWorker` in SIMONA. We use it, because we wat to know all future activation ticks. If we change the way SIMONA retrieves the next tick for primary data, we can remove this method.

src/main/java/edu/ie3/datamodel/models/timeseries/TimeSeries.java
staudtMarius: See the comment below. Since we only need it for individual time series, we can remove it for the base class.

sebastian-peter (approved):
Thanks!

PR #1443, opened 2025-09-22
Change `TimeSeries` to no longer extend `UniqueEntity`

### LINKED ISSUE #1441: Change TimeSeries implementing UniqueEntity
Currently, all time series implement the `UniqueEntity` class. The reason for this is that the individual time series should contain an uuid. Since we handle time series and unique entities seperatly, there is no need for this.

In my opinion, we can simply add the `UUID` to time series without extending `UniqueEntity`.

staudtMarius:
Resolves #1441


sebastian-peter (approved):
Looks good, thanks

PR #1461, opened 2025-10-19
Change Evcs.locationType from single element to list

### LINKED ISSUE #1460: Change `Evcs.locationType ` from single element to list
In my opinion there might be multiple use cases for an EVCS, e.g. of type `CUSTOMER_PARKING` and `WORK` at the same time.

danielfeismann:
resolves #1460


danielfeismann:
closed for #1466

PR #1470, opened 2025-11-03
Fix CosmoTimeBasedWeatherValueFactoryTest

### LINKED ISSUE #1469: `CosmoTimeBasedWeatherValueFactoryTest` does not work properly
If one changes some of the values within this test, the test is expected to fail. But it doesn't...

danielfeismann:
resolves #1469 

merge first
- #1480

### DIFF
--- a/src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
@@ -45,7 +45,7 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
     def model = factory.buildModel(data)
 
     then:
-    model.equals(expectedResults)
+    Objects.equals(model,expectedResults)
   }
 
   def "A PsdmTimeBasedWeatherValueFactory should be able to create time series values"() {
@@ -77,7 +77,7 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
     def model = factory.buildModel(data)
 
     then:
-    model.equals(expectedResults)
+    Objects.equals(model,expectedResults)
   }
 
   def "A PsdmTimeBasedWeatherValueFactory should throw FactoryException if required field is missing"() {
@@ -133,6 +133,6 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
     def model = factory.buildModel(data)
 
     then:
-    !model.equals(expectedResults)
+    !Objects.equals(model,expectedResults)
   }
 }
--- a/src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
+++ b/src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
@@ -23,7 +23,6 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
     def time = TimeUtil.withDefaults.toZonedDateTime("2019-01-01T00:00:00Z")
 
     Map<String, String> parameter = [
-      "uuid"             : "980f7714-8def-479f-baae-4deed6c8d6d1",
       "time"             : TimeUtil.withDefaults.toString(time),
       "diffuseIrradiance": "282.671997070312",
       "directIrradiance" : "286.872985839844",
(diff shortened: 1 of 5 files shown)


danielfeismann:
To reviewer: It looks to me like when testing `model == exptectedResults` the test does not use the correct equals method (of `TimeBasedValue`) but the general Object.equals(). Maybe there is a better solution as my current workaround.

danielfeismann:
@staudtMarius: After reflecting our discussion I prefer to have two separate checks of time and values instead of `model.equals(expected)` to avoid the tooltip by intellij to replace it by `==`...

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
```diff
@@ -56,7 +58,7 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
 
     Map<String, String> parameter = [
       "time"             : TimeUtil.withDefaults.toString(time),
-      "uuid"             : "980f7714-8def-479f-baae-4deed6c8d6d1",
+      "uuid"             : "c0aaf6fc-693a-458d-8d1d-d0a5093f262a",
```
staudtMarius: Why did you change the uuid? Also, since the uuid is not used to build the model, I think we schould remove it.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
```diff
@@ -78,6 +80,66 @@ class CosmoTimeBasedWeatherValueFactoryTest extends Specification {
...
+  def "A PsdmTimeBasedWeatherValueFactory should throw FactoryException if required field is missing"() {
+    given:
+    def factory = new CosmoTimeBasedWeatherValueFactory()
+    def coordinate = CosmoWeatherTestData.COORDINATE_193186
+    def time = TimeUtil.withDefaults.toZonedDateTime("2019-01-01T00:00:00Z")
+
+    // Missing 'directIrradiance' field
+    Map<String, String> parameter = [
+      "time"             : TimeUtil.withDefaults.toString(time),
+      "uuid"             : "1fc66d18-ea7d-4690-890e-9ee946c8b7b5",
```
staudtMarius: Same as above.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/TimeBasedSimpleValueFactoryTest.groovy
```diff
@@ -110,7 +110,8 @@ class TimeBasedSimpleValueFactoryTest extends Specification {
         )
 
     expect:
-    factory.buildModel(data) == expected
+    factory.buildModel(data).time == expected.time
+    factory.buildModel(data).value.toString() == expected.value.toString()
```
staudtMarius: Why did you add `toString()` here?

staudtMarius (dismissed):
Thanks for taking care of this issue. I have some small remarks.

sebastian-peter (commented):
@danielfeismann This is very, very annoying. Apparently, groovy calls `org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation#compareToWithEqualityCheck` when invoking `==`, and this in turn calls `compareTo` if the two objects implement `Comparable`. 

A secondary problem is, that apparently we don't have consistent implementations of `equals` and `compareTo` methods (see also the [documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html)). 

So, your current solution of comparing the fields of objects is the best we can do, imho, given Intellij's prevalence of suggesting replacements of `equals` with `==`.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/CosmoTimeBasedWeatherValueFactoryTest.groovy
danielfeismann: I changed them since they were not unique and used several times. But you're right, we don't need them...

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/TimeBasedSimpleValueFactoryTest.groovy
danielfeismann: Because, now, when we compare also the value they are considered as not equal.

staudtMarius:
After a short talk with @sebastian-peter, the best way would be to use `Objects.equals()` to write this test.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/TimeBasedSimpleValueFactoryTest.groovy
danielfeismann: But maybe this should be fixed in the equals method of `EnergyPriceValue` ? #1479

sebastian-peter (approved):
Looks good, thanks

PR #1473, opened 2025-11-10
Adapt PSDM to parse simonaMarkovLoad JSON models

### LINKED ISSUE #1472: Implement Markov-based PowerValueSource
Introduce a Markov-based implementation of PowerValueSource that consumes JSON model files generated by the simonaMarkovLoad project.

PhilippSchmelter:
Closes #1472


build.gradle
```diff
@@ -105,6 +105,8 @@ dependencies {
   implementation 'commons-io:commons-io:2.21.0' // I/O functionalities
   implementation 'commons-codec:commons-codec:1.20.0' // needed by commons-compress
   implementation 'org.apache.commons:commons-compress:1.28.0' // I/O functionalities
+
+  implementation 'com.fasterxml.jackson.core:jackson-databind:2.17.2'
```
staudtMarius: This should be changed to the newest version of jackson2 (`2.20.1`) or the jackson3. But the change to jackson3 can be done in another PR.

src/main/java/edu/ie3/datamodel/io/connectors/JsonFileConnector.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Opens a buffered reader for the given JSON file, using UTF-8 decoding.
+   *
+   * @param filePath relative path without ending
+   * @return buffered reader referencing the JSON file
+   */
+  public BufferedReader initReader(Path filePath) throws FileNotFoundException {
+    InputStream inputStream = openInputStream(filePath);
+    return new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8), 16384);
+  }
```
staudtMarius: Is this necessary? I think this could be removed, since we are only using the input stream to read in the JSON file.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,318 @@
...
+      while (fields.hasNext()) {
+        Map.Entry<String, JsonNode> entry = fields.next();
+        config.put(entry.getKey(), entry.getValue().asText());
+      }
+    }
+    return new Generator(name, version, config);
+  }
+
+  default TimeModel parseTimeModel(JsonNode timeNode) {
+    int bucketCount = requireInt(timeNode, "bucket_count");
```
staudtMarius: I think naming the method `acquiere...` or `extract...` would be better.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,318 @@
...
+    Parameters.TransitionParameters transitions =
+        new Parameters.TransitionParameters(
+            parametersNode.path("transitions").path("empty_row_strategy").asText(""));
+    if (transitions.emptyRowStrategy().isEmpty()) {
+      transitions = null;
+    }
+
+    JsonNode gmmNode = parametersNode.path("gmm");
+    Parameters.GmmParameters gmm =
+        gmmNode.isMissingNode() || gmmNode.isNull() || gmmNode.size() == 0
```
staudtMarius: Could be improved by using `gmmNode.isEmpty()`.

src/main/java/edu/ie3/datamodel/io/naming/EntityPersistenceNamingStrategy.java
```diff
@@ -61,7 +61,7 @@ public class EntityPersistenceNamingStrategy {
    * profile is accessible via the named capturing group "profile", the uuid by the group "uuid"
    */
   private static final String LOAD_PROFILE_TIME_SERIES =
-      "lpts_(?<profile>[a-zA-Z]{1,11}[0-9]{0,3})";
+      "(?:lpts|markov)_(?<profile>[a-zA-Z]{1,11}[0-9]{0,3})";
```
staudtMarius: This could lead to a problem if someone is using `lpts_h0.csv` and `markov_h0.json`. We should add some explanation to the docs that the profile names need to be unique.

staudtMarius (changes requested):
Here are some first comments. But we should talk about the source again. With the current implementation we split the markov power source into two source classes, while the other power sources are implemented with one class.

Two more things from my side:
- Could you add more javadocs and comments in the code to improve the readability of the code?
- Could you add some information for readTheDocs?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovLoadModelFactory.java
```diff
@@ -0,0 +1,83 @@
...
+    ZonedDateTime generatedAt = parseTimestamp(requireText(root, "generated_at"));
+    Generator generator = parseGenerator(requireNode(root, "generator"));
+    TimeModel timeModel = extractTimeModel(requireNode(root, "time_model"));
+    ValueModel valueModel = parseValueModel(requireNode(root, "value_model"));
+    Parameters parameters = parseParameters(root.path("parameters"));
+
+    JsonNode dataNode = requireNode(root, "data");
+    TransitionData transitionData =
+        parseTransitions(dataNode, timeModel.bucketCount(), valueModel.discretization().states());
+    GmmBuckets gmmBuckets = parseGmmBuckets(requireNode(dataNode, "gmms"));
```
staudtMarius: Does it makes sense to have the method `parseGmmBuckets` to return an optional directly?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+import tools.jackson.databind.JsonNode;
+
+/**
+ * Shared parsing helpers for Markov model JSON documents. This is intentionally package-private as
+ * it is only meant to be reused across factory implementations in this package.
+ */
+interface MarkovModelParsingSupport {
+
+  default Generator parseGenerator(JsonNode generatorNode) {
+    String name = requireText(generatorNode, "name");
```
staudtMarius: The methods are still names as `require*`. I would prefer if you could rename them `acquiere*` or `extract*`.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+      throw new FactoryException(
+          "Discretization thresholds_right must contain "
+              + Math.max(0, states - 1)
+              + " entries for "
+              + states
+              + " states, but found "
+              + thresholds.size());
+    }
+    ValueModel.Discretization discretization =
+        new ValueModel.Discretization(states, List.copyOf(thresholds));
```
staudtMarius: Can't you add `thresholds` here directly?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+    List<GmmBuckets.GmmBucket> buckets = new ArrayList<>();
+    for (JsonNode bucketNode : bucketsNode) {
+      JsonNode statesNode = bucketNode.get("states");
+      if (statesNode == null || !statesNode.isArray()) {
+        throw new FactoryException("Each GMM bucket must contain an array 'states'");
+      }
+      List<Optional<GmmBuckets.GmmState>> states = new ArrayList<>();
+      for (JsonNode stateNode : statesNode) {
+        if (stateNode == null || stateNode.isNull()) {
+          states.add(Optional.empty());
```
staudtMarius: Why are you adding an empty optional here and another optional later?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+          continue;
+        }
+        List<Double> weights = readDoubleArray(stateNode, "weights");
+        List<Double> means = readDoubleArray(stateNode, "means");
+        List<Double> variances = readDoubleArray(stateNode, "variances");
+        states.add(Optional.of(new GmmBuckets.GmmState(weights, means, variances)));
+      }
+      buckets.add(new GmmBuckets.GmmBucket(List.copyOf(states)));
```
staudtMarius: Same question as above: Why are you using `List.copyOf`?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+    return value.asString();
+  }
+
+  default double requireDouble(JsonNode node, String field) {
+    JsonNode value = node.get(field);
+    if (value == null || value.isMissingNode() || value.isNull()) {
+      throw new FactoryException("Missing field '" + field + "'");
+    }
+    if (!value.isNumber()) {
+      throw new FactoryException("Field '" + field + "' must be numeric");
```
staudtMarius: `... must be a double`?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+      throw new FactoryException(
+          "Transition values provided only " + bucketIndex + " buckets. Expected " + buckets);
+    }
+    return values;
+  }
+
+  default void fillBucket(
+      double[][][] values, JsonNode bucketNode, int bucketIndex, int stateCount) {
+    if (bucketIndex >= values.length) {
+      throw new FactoryException("More transition buckets present than specified in shape");
```
staudtMarius: This could be checked once in `parseTransitionValues` by comparing `buckets` and `valuesNode` size.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,324 @@
...
+  }
+
+  default List<Double> readDoubleArray(JsonNode node, String field) {
+    JsonNode arrayNode = node.get(field);
+    if (arrayNode == null || !arrayNode.isArray()) {
+      throw new FactoryException("Field '" + field + "' must be an array");
+    }
+    List<Double> values = new ArrayList<>();
+    arrayNode.forEach(element -> values.add(element.asDouble()));
+    return List.copyOf(values);
```
staudtMarius, line 327: Same as above.

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public JsonMarkovProfileSource(
+      JsonDataSource dataSource,
+      FileLoadProfileMetaInformation metaInformation,
+      MarkovLoadModelFactory factory) {
+    this.dataSource = Objects.requireNonNull(dataSource, "dataSource");
+    this.metaInformation = Objects.requireNonNull(metaInformation, "metaInformation");
+    this.factory = Objects.requireNonNull(factory, "factory");
+    this.profile = new MarkovPowerProfile(metaInformation.getProfileKey());
```
staudtMarius: Why are you adding a `MarkovPowerProfile`?

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private JsonNode readModelTree() throws SourceException {
+    Path filePath = metaInformation.getFullFilePath();
+    try (InputStream inputStream = dataSource.initInputStream(filePath)) {
+      return objectMapper.readTree(inputStream);
+    } catch (IOException e) {
+      throw new SourceException("Unable to read Markov model JSON from '" + filePath + "'.", e);
+    }
+  }
```
staudtMarius: Most of this metho should be moved to the data source. Then you could call it via the method `getSourceFields(Path)`.

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private static Set<String> collectFieldNames(JsonNode node) {
```
staudtMarius: This method should also be moved to the data source.

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
```diff
@@ -0,0 +1,178 @@
...
+    Set<String> fields = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+    collectFields("", node, fields);
+    return fields;
+  }
+
+  private static void collectFields(String prefix, JsonNode node, Set<String> collector) {
```
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,536 @@
...
+  }
+
+  /** Returns the maximum power from the model's normalization configuration. */
+  public Optional<ComparableQuantity<Power>> getMaxPower() {
+    return Optional.of(maxPowerFromModel);
+  }
+
+  /** Markov models do not define an energy scaling factor. */
+  public Optional<ComparableQuantity<Energy>> getProfileEnergyScaling() {
+    return Optional.empty();
```
staudtMarius, line 204: Please add the information that energy scaling is not supported by the markov based load model to the docs.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,536 @@
...
+    }
+
+    private double nextGaussian(SplittableRandom rng) {
+      double u1 = Math.max(Double.MIN_VALUE, rng.nextDouble());
+      double u2 = rng.nextDouble();
+      return Math.sqrt(-2.0d * Math.log(u1)) * Math.cos(2.0d * Math.PI * u2);
+    }
+  }
+
+  private final class MarkovSupplier {
```
staudtMarius: Is this class necessary? You could move some code to the the method `getValueSupplier` and the private method could just be private methods of the source.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovPowerProfile.java
```diff
@@ -0,0 +1,37 @@
...
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.models.profile.markov;
+
+import edu.ie3.datamodel.models.profile.PowerProfile;
+import edu.ie3.datamodel.models.profile.PowerProfileKey;
+import java.util.Objects;
+
+/** Simple {@link PowerProfile} implementation for Markov-based load models. */
+public record MarkovPowerProfile(PowerProfileKey key) implements PowerProfile {
```
staudtMarius: In my opinion, this class is unnecessary. It just duplicates the behavior of `PowerProfileKey`.

staudtMarius (changes requested):
Here are some more remarks from my side.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: changed to extract

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: refactored to use the existing readDoubleArray helper, which already returns an immutable list.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: You are right this was kind of confusing. I just removed all optionals which were in collections because null already represents absence.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
staudtMarius: Does it have to be an immutable list? If the have no setters, there shouldn't be an issue with having mutable collections.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: changed to Collections.unmodifiableList so no copy is made and an unmodifiable view is provided (read only). Now that we dont use Optionals anymore we work with nulls, which are not allowed in List.copyOf. So this is the new clean approach

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: i changed the message to double

PhilippSchmelter: might be more precise and less confusing

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: now we validate once upfront following failfast

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: List.copyOf is fine here since doubles cant be null. We have the option to use a stream but we keep forEach for readability.

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
PhilippSchmelter: turns out MarkovPowerProfile was just sitting there looking pretty without doing anything :D
i kind of lost focus somewhere between thinking and coding. i removed it

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: you are right it doesnt have to be immutable since  ValueModel.Discretization is a record with no setters. we can keep it like that tho
its a defensive approach

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: good point! i moved that computation logic into private methods of the Model and simplified getValueSupplier to a lambda

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -255,162 +258,121 @@ private double nextGaussian(SplittableRandom rng) {
...
-          throw new IllegalArgumentException("Previous state out of bounds: " + state);
+  private double[] sanitizeDistribution(int bucket, double[] row) {
+    // Ignore invalid probabilities and states without GMM data, then renormalize.
+    double[] sanitized = new double[stateCount];
+    double sum = 0d;
+    for (int state = 0; state < stateCount; state++) {
+      double sanitizedValue = 0d;
+      if (state < row.length) {
+        double value = row[state];
+        if (value > 0d && !Double.isNaN(value) && gmmStates[bucket][state] != null) {
```
staudtMarius: The check `!Double.isNaN(value)` cannot be false, because a java primitive type cannot be null. Therefore, you can remove this check.

staudtMarius (changes requested):
One last comment from my side.

@sebastian-peter Do you want to have a look?

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: i removed it

docs/readthedocs/io/csvfiles.md
```diff
@@ -160,6 +160,15 @@ The following keys are supported until now:
...
 
 The following profiles are supported until now:
+
+Note:
+Load profile keys must be unique across all load-profile sources. Do not use the same profile key
+for both CSV (lpts_*) and Markov JSON (markov_*), e.g. avoid having both lpts_h0.csv and
+markov_h0.json.
+
+Markov-based load models (`markov_*`) do not support energy scaling. Calls to
+`getProfileEnergyScaling()` will always return `Optional.empty()`.
```
sebastian-peter, line 160: Hmm, why not? We should be able to figure out an average sum that a Markov model produces over a year... Maybe something for a later PR?

src/test/groovy/edu/ie3/datamodel/io/factory/markov/MarkovLoadModelFactoryTest.groovy
```diff
@@ -0,0 +1,185 @@
...
+            "states": 2,
+            "thresholds_right": [0.5]
+          }
+        },
+        "parameters": {
+          "transitions": { "empty_row_strategy": "fill" },
+          "gmm": {
+            "value_col": "p",
+            "verbose": 1,
+            "heartbeat_seconds": 5
```
sebastian-peter: I tried to find out what a heart beat would be in the context of GMM but couldn't, maybe you can enlighten me?

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter, line 1: We definitely need documentation for at least the model (basically the records) and the methods (besides simple ones like getters) as well. Could you add javadoc for these? I realize it's a bit of work, but it could be very beneficial to the project, I think

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,498 @@
...
+    double sample = rng.nextDouble();
+    double cumulative = 0d;
+    for (int i = 0; i < distribution.length; i++) {
+      cumulative += distribution[i];
+      if (sample <= cumulative) {
+        return i;
+      }
+    }
+    return distribution.length - 1;
+  }
```
sebastian-peter: Almost complete duplicate of `GmmStateData.drawComponent`

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,498 @@
...
+    private static double[] toArray(List<Double> values) {
+      double[] array = new double[values.size()];
+      for (int i = 0; i < values.size(); i++) {
+        array[i] = values.get(i);
+      }
+      return array;
+    }
+
+    private double sample(SplittableRandom rng) {
+      int component = drawComponent(rng.nextDouble());
```
sebastian-peter: It looks like we're sampling a random value from a random distribution here...? Shouldn't this be drawing a random value from a specific distribution, namely the one of the current state?

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter, line 1: In general, the arrangement of methods seems a bit unstructured to me... Following the trail of method calls, we're jumping up and down in the file without apparent reason. Could you make this a bit more orderly? Currently, there's no way for me to figure out if this is doing what it is supposed to or not...

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,498 @@
...
+    double clamped = clamp01(normalizedValue);
+    ComparableQuantity<Power> range = maxPowerFromModel.subtract(minPowerFromModel);
+    return minPowerFromModel.add(range.multiply(clamped)).asType(Power.class);
+  }
+
+  private double clamp01(double value) {
+    if (value < 0d) return 0d;
+    if (value > 1d) return 1d;
+    return value;
+  }
```
sebastian-peter: Why do we need to clamp values? Not a trick question, just trying to understand...

sebastian-peter (changes requested):
Some initial thoughts and questions... Would be great if we could make an appointment where you walk me through how this model works

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,498 @@
...
+    if (!"kW".equalsIgnoreCase(reference.unit())) {
+      throw new IllegalArgumentException(
+          "Unsupported reference power unit '" + reference.unit() + "'. Only kW is supported.");
+    }
+    return Quantities.getQuantity(reference.value(), StandardUnits.ACTIVE_POWER_IN);
+  }
+
+  private record StepResult(int nextState, double normalizedValue) {}
+
+  private static final class GmmStateData {
```
sebastian-peter: Another question would be: Why is there `GmmStateData` and `GmmBuckets.GmmState`, which seem to hold the same data, just one with arrays and one with lists?

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: youre right
refactored both into a shared helper!

docs/readthedocs/io/csvfiles.md
PhilippSchmelter: I want to keep this PR as simple as possible. I know there will be some more work to do in future PRs. Technically we have everything to do that

src/test/groovy/edu/ie3/datamodel/io/factory/markov/MarkovLoadModelFactoryTest.groovy
PhilippSchmelter: it is not a GMM concept. it is a python training parameter. the generator uses it to detect hung worker processes during parallel gmm fitting. The entire parameters.gmm block is just the generators config. none of it is used on java side. i could remove it on the python side or just ignore it

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: i know it looks confusing at first glance but it is correct. sample() is always called on a specific GmmStateData Instance.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovLoadModelFactory.java
```diff
@@ -0,0 +1,50 @@
...
+  public MarkovLoadModelFactory() {
+    super(MarkovLoadModel.class);
+  }
+
+  /** Builds a {@link MarkovLoadModel} from a parsed JSON tree. */
+  @Override
+  protected MarkovLoadModel buildModel(MarkovModelData data) {
+    JsonNode root = data.getRoot();
+    String schema = extractText(root, "schema");
+    ZonedDateTime generatedAt = parseTimestamp(extractText(root, "generated_at"));
```
staudtMarius: Please use the names defined in `FieldNamingStrategy` here.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,328 @@
...
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeParseException;
+import java.util.*;
+import tools.jackson.databind.JsonNode;
+
+/** Shared parsing helpers for Markov model JSON documents. */
+interface MarkovModelParsingSupport {
+
+  default Generator parseGenerator(JsonNode generatorNode) {
+    String name = extractText(generatorNode, "name");
```
staudtMarius: Please use the names defined in FieldNamingStrategy here. Same for all other places.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,328 @@
...
+      }
+    }
+    return new Generator(name, version, config);
+  }
+
+  /** Extracts the time model block. */
+  default TimeModel extractTimeModel(JsonNode timeNode) {
+    int bucketCount = extractInt(timeNode, "bucket_count");
+    if (bucketCount <= 0) {
+      throw new FactoryException("time_model.bucket_count must be positive");
```
staudtMarius: Zero is also positive. Maybe change this to `... must be greater than 0`. Also, could you add a `.` at the end of the exception messages?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,328 @@
...
+    }
+    String formula = extractNode(timeNode, "bucket_encoding").path("formula").asString("");
+    if (formula.isEmpty()) {
+      throw new FactoryException("Missing bucket encoding formula");
+    }
+    int samplingInterval = extractInt(timeNode, "sampling_interval_minutes");
+    if (samplingInterval <= 0) {
+      throw new FactoryException("time_model.sampling_interval_minutes must be positive");
```
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,328 @@
...
+    ValueModel.Normalization normalization =
+        new ValueModel.Normalization(
+            normalizationMethod,
+            parsePowerReference(normalizationNode, "max_power"),
+            parsePowerReference(normalizationNode, "min_power"));
+
+    JsonNode discretizationNode = extractNode(valueNode, "discretization");
+    int states = extractInt(discretizationNode, "states");
+    if (states <= 0) {
+      throw new FactoryException("value_model.discretization.states must be positive");
```
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/io/naming/timeseries/LoadProfileMetaInformation.java
```diff
@@ -27,6 +35,18 @@ public PowerProfileKey getProfileKey() {
     return profileKey;
   }
 
+  private static PowerProfileKey.Type parseType(String type) {
```
staudtMarius: This could be moved to `PowerProfileKey`.

src/main/java/edu/ie3/datamodel/io/naming/EntityPersistenceNamingStrategy.java
```diff
@@ -57,10 +57,11 @@ public class EntityPersistenceNamingStrategy {
 
   /**
    * Regex to match the naming convention of a file for a repetitive load profile time series. The
```
staudtMarius: Could you adapt the first sentence? The time series part does not fit markov profiles.

src/main/java/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSource.java
```diff
@@ -0,0 +1,140 @@
...
+  public JsonMarkovProfileSource(
+      JsonDataSource dataSource, FileLoadProfileMetaInformation metaInformation) {
+    this(dataSource, metaInformation, new MarkovLoadModelFactory());
+  }
+
+  public JsonMarkovProfileSource(
+      JsonDataSource dataSource,
+      FileLoadProfileMetaInformation metaInformation,
+      MarkovLoadModelFactory factory) {
+    this.dataSource = Objects.requireNonNull(dataSource, "dataSource");
```
staudtMarius: Could you remove the `requireNonNull` calls here?

src/main/java/edu/ie3/datamodel/io/source/PowerValueSource.java
```diff
@@ -72,4 +78,53 @@ sealed interface PowerValueIdentifier permits PowerValueSource.TimeSeriesInputVa
...
+   */
+  record MarkovIdentifier(
+      ZonedDateTime time,
+      OptionalInt previousState,
+      OptionalDouble initialNormalizedValue,
+      long randomSeed)
+      implements PowerValueIdentifier {
+
+    public MarkovIdentifier {
+      Objects.requireNonNull(time, "time");
```
staudtMarius: Same here.

src/test/groovy/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSourceTest.groovy
```diff
@@ -0,0 +1,221 @@
...
+    MarkovLoadModel modelFirst = source.getModel()
+    MarkovLoadModel modelSecond = source.getModel()
+
+    then:
+    modelFirst.is(modelSecond) // cached instance reused
+    modelFirst.schema() == "markov.load.v1"
+    noExceptionThrown()
+
+    when: "validation is executed on the same file"
+    source.validate()
```
staudtMarius: Could you first test the validation and then the model?

src/test/groovy/edu/ie3/datamodel/io/source/json/JsonMarkovProfileSourceTest.groovy
```diff
@@ -0,0 +1,221 @@
...
+    then:
+    source.getProfileKey().getValue() == "profile1"
+    source.getMaxPower().isPresent()
+    source.getMaxPower().get().to(StandardUnits.ACTIVE_POWER_IN).value.doubleValue() == 10d
+    output.value().isPresent()
+    output.nextState() == 1
+    output.value().get().p.get().to(StandardUnits.ACTIVE_POWER_IN).value.doubleValue() == 5.25d
+  }
+
+  private static String validModelJson() {
```
staudtMarius: Maybe you could move some common json test methods to an abstract base class.

staudtMarius (changes requested):
Here are some comments from my side.

staudtMarius (dismissed):
LGTM

staudtMarius (dismissed):

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+  public Optional<ZonedDateTime> getNextTimeKey(ZonedDateTime time) {
+    return Optional.ofNullable(time).map(t -> t.plusMinutes(samplingIntervalMinutes));
+  }
+
+  /** Returns the maximum power from the model's normalization configuration. */
+  public Optional<ComparableQuantity<Power>> getMaxPower() {
+    return Optional.of(maxPowerFromModel);
+  }
+
+  /** Not implemented yet. Future feature */
```
sebastian-peter: Please create an issue for this, so that this doesn't get lost

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+          + ", encoding="
+          + encoding
+          + ", values="
+          + Arrays.deepToString(values)
+          + "]";
+    }
+  }
+
+  /** Per-bucket Gaussian Mixture Models. State entries may be {@code null}. */
+  public record GmmBuckets(List<GmmBucket> buckets) {
```
sebastian-peter, line 408: I think you could reduce code size (which is always good, I think) if you merged GmmBuckets/GmmBucket/GmmState and GmmStateData. I guess an argument for keeping it like it is now is that GmmBuckets etc. show the structure of the data, which might be helpful for understanding the parsing code... But with some commentary on `private final GmmStateData[][] gmmStates` and within `buildGmmStates`, that should be achievable as well?

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter: After thinking it through again, makes sense now

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+
+    /**
+     * Normalization range used to map between normalized values and physical power. The optional
+     * wrapping reflects the JSON shape, both bounds are required at construction time.
+     */
+    public record Normalization(
+        String method, Optional<PowerReference> maxPower, Optional<PowerReference> minPower) {
+
+      /** A value/unit pair from the JSON normalization block. Only {@code "kW"} is supported. */
+      public record PowerReference(double value, String unit) {}
```
sebastian-peter: You could also use `ComparableQuantity<Power>` instead, since we're using this everywhere in PSDM and this is called the Markow**Load**Model... So it's fine to expect a power here, I think.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    /** State-bin layout and right-edge thresholds. */
+    public record Discretization(int states, List<Double> thresholdsRight) {}
```
sebastian-peter, line 341: Don't know if this could be transformed into some general state limit description... Maybe depends on the other issues I've raised and if they're actual issues

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+    seed = 31 * seed + bucket;
+    seed = 31 * seed + state;
+    long slot = input.time().toInstant().toEpochMilli() / (samplingIntervalMinutes * 60_000L);
+    return 31 * seed + slot;
+  }
+
+  /** Draws the next state from the transition row and samples the corresponding GMM. */
+  private StepResult simulateStep(int bucket, int currentState, SplittableRandom rng) {
+    int nextStateIndex = drawWeighted(transitions[bucket][currentState], rng);
+    double normalized = sampleNormalizedValue(bucket, nextStateIndex, rng);
```
sebastian-peter, line 276: You're sampling a value here between 0 and 1... Are 0 and 1 the state limits of the state that we sampled for, or the overall minimum and maximum power of the model?

(Note for myself: clamping)

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,376 @@
...
+    if (value == null || value.isMissingNode() || value.isNull()) {
+      throw new FactoryException("Missing field '" + field + "'.");
+    }
+    if (!value.isString()) {
+      throw new FactoryException("Field '" + field + "' must be textual.");
+    }
+    return value.asString();
+  }
+
+  default double extractDouble(JsonNode node, String field) {
```
sebastian-peter: This could be merged with `extractDoubleValue`

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,376 @@
...
+  default OptionalInt optionalInt(JsonNode node, String field) {
+    JsonNode value = node.get(field);
+    if (value == null || value.isNull()) return OptionalInt.empty();
+    if (!value.canConvertToInt()) {
+      throw new FactoryException("Field '" + field + "' must be an integer.");
+    }
+    return OptionalInt.of(value.asInt());
+  }
+
+  default int[] parseTransitionShape(JsonNode transitionsNode) {
```
sebastian-peter: Some JavaDoc on the return type would be great here...

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
```diff
@@ -0,0 +1,592 @@
...
+
+    if (!this.maxPowerFromModel.isGreaterThan(this.minPowerFromModel)) {
+      throw new IllegalArgumentException(
+          "Markov model normalization has non-positive range (max <= min).");
+    }
+
+    validateTransitionDimensions();
+  }
+
+  private void validateTransitionDimensions() {
```
sebastian-peter: Maybe all validation functionality could be in one place (`MarkovModelParsingSupport`, where there's already `validateTransitionShape`)?

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
```diff
@@ -0,0 +1,376 @@
...
+              extractDoubleValue(
+                  rowNode.get(c),
+                  jsonField(MARKOV_TRANSITION_VALUES) + "[" + b + "][" + r + "][" + c + "]");
+        }
+      }
+    }
+    return values;
+  }
+
+  default JsonNode extractTransitionBucket(JsonNode valuesNode, int bucket, int stateCount) {
```
sebastian-peter: This reads more like a validation method than an extraction method, same for the other methods below. But maybe that's the way to go? So maybe `validateTransitionShape` could be integrated into `parseTransitionShape`?

Also (and this is a bit pedantic, I know), maybe you could name parameters like `stateCount`, that are just used for validation, something like `expectedStateCount`?

sebastian-peter (changes requested):
Some more questions/comments from my side. Overall: great work in improving the structure of your code, this was a breeze in going through it.

src/main/java/edu/ie3/datamodel/io/factory/markov/MarkovModelParsingSupport.java
PhilippSchmelter: Good point, I merged the two methods. `parseTransitionShape` returned the shape but after
`validateTransitionShape` those values are fixed anyway (`buckets == expectedBucketCount` and
`rows == columns == stateCount`), so the returned array didn't carry any information. It's now a
single `validateTransitionShape(transitionsNode, expectedBucketCount, expectedStateCount)` that
reads and checks the shape, and `parseTransitionValues` takes the counts from the caller directly.
As a side effect the five consecutive int parameters are gone as well, which were easy to mix up.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: Bundled the model side checks in a new package private class `MarkovModelValidation`. I kept them
out of `MarkovModelParsingSupport`, since that one is package private in `io.factory.markov` and
validates the JSON document, while these are model invariants that have to stay in the constructor.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: I merged them. If i remember correctly, someday it already was merged but someone commented that two drawWeighted methods could be merged and thats how we came to the current state. Now i found a new possibility how to do it with just one drawWeighted method:
 `drawWeighted` is used for the transition rows (`double[]`) and the GMM
weights (`List<Double>`), so it now takes an index accessor instead of an array. That avoids having
two copies of it, but tell me if you find it too indirect.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: 0 and 1 are the model wide normalization range, not the limits of the state. The input data is
normalized to [0, 1], so we clamp the Gaussian tails to that. Made that explicit in the javadoc.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
PhilippSchmelter: `thresholds_right` already is that state limit description, just written as shared edges: state i
covers [t(i-1), t(i)), with 0 and 1 from the model wide normalization as the outer bounds. Writing
it out as explicit per state intervals would duplicate every edge and move away from the JSON,
which gives us the thresholds in exactly this form.

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter: #1696

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter: I mean, one could as further why we're using arrays for the transition weights and lists for the GMM data... ;D But I'm fine with it this way

src/main/java/edu/ie3/datamodel/models/profile/markov/MarkovLoadModel.java
sebastian-peter: I think I meant something else with my comment, but it doesn't matter that much

sebastian-peter (dismissed):
Thanks for all the extra work on this! I had one final question on my mind, but let's not get into more refactoring ;)

https://github.com/ie3-institute/PowerSystemDataModel/pull/1473#discussion_r3923075172

sebastian-peter (approved):

PR #1480, opened 2025-11-18
Fix equals method of EnergyPriceValue

### LINKED ISSUE #1479: Fix `EnergyPriceValue.equals()` method

danielfeismann:
resolves #1479

### DIFF
--- a/src/main/java/edu/ie3/datamodel/models/value/EnergyPriceValue.java
+++ b/src/main/java/edu/ie3/datamodel/models/value/EnergyPriceValue.java
@@ -33,8 +33,8 @@ public class EnergyPriceValue implements Value {
     if (o == null || getClass() != o.getClass()) return false;
     EnergyPriceValue that = (EnergyPriceValue) o;
 
-    return price.getValue().doubleValue() == that.price.getValue().doubleValue()
-        && price.getUnit().equals(that.price.getUnit());
+    return Objects.equals(price.getValue(), that.price.getValue())
+        && Objects.equals(price.getUnit(), that.price.getUnit());
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


src/main/java/edu/ie3/datamodel/models/value/EnergyPriceValue.java
```diff
@@ -32,12 +32,14 @@ public boolean equals(Object o) {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     EnergyPriceValue that = (EnergyPriceValue) o;
-    return Objects.equals(price, that.price);
+
+    return price.getValue().doubleValue() == that.price.getValue().doubleValue()
+        && price.getUnit().equals(that.price.getUnit());
```
staudtMarius: Does it makes sense to use `Objects.equals(price.getUnit(), that.price.getUnit())` instead?

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/TimeBasedSimpleValueFactoryTest.groovy
```diff
@@ -110,7 +110,7 @@ class TimeBasedSimpleValueFactoryTest extends Specification {
         )
 
     expect:
-    factory.buildModel(data) == expected
+    factory.buildModel(data).equals(expected)
```
staudtMarius: I think it would be better to use `Objects.equals()` for the tests, since the intellij won't suggest switching to `==`.

staudtMarius (changes requested):
Looks good to me, Just two small comments.

staudtMarius (approved):
LGTM.

PR #1510, opened 2026-01-05
Requiring thermal bus field in thermal house and storage input

### LINKED ISSUE #1509: Adding `thermalBus` as required field in house and storage input

sebastian-peter:
Closes #1509


staudtMarius (approved):
LGTM.

danielfeismann:
Thanks for fixing this. However, I think it would be a good idea to include this in the `ThermalValidationUtils` too. What do you think? @sebastian-peter

sebastian-peter:
> Thanks for fixing this. However, I think it would be a good idea to include this in the `ThermalValidationUtils` too. What do you think? @sebastian-peter

I think this is not really necessary, as this is caught by the parser already (which should be fixed by #1512). On the contrary, I think it's not possible to create a thermal model that is successfully parsed and that would fail such validation. @danielfeismann

Issue #976, opened 2024-01-03
Check whether `FactoryData` and subclasses can be eliminated


sebastian-peter:
As another successive issue of #969, there might be a way to eliminate `FactoryData` and all its subclasses (`EntityData` etc.) by creating `AssetInput`s and similar entities directly. 
To make this work, the new constructors in the form of `AssetInputEntityData(EntityData entityData, OperatorInput operator)` (but of course adapted for `AssetInput`s) could be used to iteratively create Inputs from the simplest `UniqueEntity` to the desired input type.

A side effect of this would be that error messages would be easier to create, since there is no way of ensuring that a `FactoryData` also includes UUID and id.

staudtMarius:
@sebastian-peter To do this we'd have to change classes like `AssetInput` and `UniqueInput` from being abstract. This should be done after #1044.

sebastian-peter:
> @sebastian-peter To do this we'd have to change classes like `AssetInput` and `UniqueInput` from being abstract. This should be done after #1044.

Yes, and I'm not really sure if that'd be smart. We should think this through beforehand.

staudtMarius:
@sebastian-peter If you want, we could talk about this next week.

PR #1512, opened 2026-01-06
Fixed handling of erroneous field values in `EntitySource.enrichWithDefault()`

### LINKED ISSUE #1511: No exception is thrown if UUID in `controlling_em` does not link to valid EM
I think the way `enrichWithDefault` works can be improved. Currently, it just swallows any error that occurs in `extractFunction` and the user has no clue why unexpected things happen. 

https://github.com/ie3-institute/PowerSystemDataModel/blob/e56725cc8fff20965e17ec6bf48f27cc69f5e5c5/src/main/java/edu/ie3/datamodel/io/source/EntitySource.java#L248-L259

This also affects the operator assignment in `AssetEntitySource`, as it also uses `enrichWithDefault`.

The way this should work is this, in my opinion:
- any non-empty string that is not a valid UUID -> error
- valid UUID, but there is no matching entity -> error
- empty string -> only then use default value

staudtMarius:
Resolves #1511


sebastian-peter (approved):
Looks ok - although I have to say that `EntitySource` has become so complex now that it's hard to understand (for example: `enrichFunction` returns a function take takes a `Pair`, which should contain the entity data as key and the entity or entities to enhance with as value. This pair is generated in `enrich` etc.). I don't see an easy way to improve this, besides enhancing javadoc (the description "Method to build an enrich function." is used everywhere, although the functions do very different things, for example).

staudtMarius:
The complexity could be reduced with #976.

PR #1547, opened 2026-02-11
Removed `ThermalUnits` (duplication to `ThermalGrids`)

### LINKED ISSUE #874: `ThermalUnits` vs `ThermalGrids`
Currently we have ThermalUnits and ThermalGrids which both represent collection of ThermalHouses and ThermalStorages. Afaik we only use and need the grids so we should discuss getting rid of the units.

danielfeismann:
resolves #874


sebastian-peter (approved):
Thanks for the cleaning! Didn't even realize that `ThermalUnits` was never even used anywhere

danielfeismann:
Me neither. Naming of classes is still not perfect. We have still ThermalUnitsResult, ThermalUnitInput... It would be great if they would match semantical together with ThermalGrid as we have that in other classes. However, ThermalGrid suites to me much better as collection for the houses and storages than this one here. Maybe someone has any good idea on this, we can further improve.

PR #1571, opened 2026-03-09
Updated attribute units for standard lines types

### LINKED ISSUE #1483: Emphasize attribute units for standard lines types
Don't know what the best solution would be, but somehow, I think, we should mention in the [standard line type section](https://powersystemdatamodel.readthedocs.io/en/latest/models/input/grid/line.html#standard-line-types) what units the columns have. I think not everybody will make the connection to the [type table](https://powersystemdatamodel.readthedocs.io/en/latest/models/input/grid/line.html#type-model) at the top.

Specifically, it should be more obvious that b, g, r, x are specified as Ohm/km and µS / km.

Same thing applies to standard transfomers, although less urgent

kkarthik11:
Resolves #1483 attribute units for standard lines types and changelog update


danielfeismann (changes requested):
Thanks for tackling this issue. Could you please do the same also for the two- and three-winding transformers? And I would prefer to have the units in square brackets not normal brackets, but thats just my opinion.

docs/readthedocs/models/input/grid/transformer2w.md
```diff
@@ -24,39 +24,39 @@ As obvious, the parameter can be used in T- as in 𝜋-equivalent circuit repres
      - Human readable identifier
```
danielfeismann, line 24: please roll back this for the type model table but add the units in square brackets for the data for the standard asset table below

docs/readthedocs/models/input/grid/line.md
```diff
@@ -24,27 +24,27 @@ Representation of an AC line.
      - Human readable identifier
 
    * - r
-     -  Ω / km
+     -  [Ω / km]
```
danielfeismann: please roll back this for the type model table. We only need the units in brackets for the data for the standard asset table below

docs/readthedocs/models/input/grid/transformer3w.md
```diff
@@ -41,67 +41,67 @@ All impedances and admittances are given with respect to the higher voltage side
      - Human readable identifier
 
    * - rScA
-     - Ω
+     - [Ω]
```
danielfeismann: same as for 2w-transformer, rollback for type model table but add for the standard asset data table below

danielfeismann (changes requested):

docs/readthedocs/models/input/grid/line.md
```diff
@@ -316,13 +316,13 @@ Some standard cables.
    
    
    * - uuid
-     - b
-     - g
-     - iMax
+     - b [µS / km]
+     - g [S / km]
```
danielfeismann:
```suggestion
     - g [µS / km]
```

docs/readthedocs/models/input/grid/transformer2w.md
```diff
@@ -48,7 +48,7 @@ As obvious, the parameter can be used in T- as in 𝜋-equivalent circuit repres
      - Rated voltage at higher voltage terminal
 
    * - vRatedB
-     - kV
+     - pkV
```
danielfeismann:
```suggestion
     - kV
```

danielfeismann (approved):
Thanks for improving the docs, i made two minor corrections.

PR #1575, opened 2026-03-18
Refactor validation of value classes.

### LINKED ISSUE #1558: Refactor source validation
Currently, the source validation is part of the factories, but it should be moved to the data source. This is necessary to work on #976.

### LINKED ISSUE #1560: Refactor validation of value classes

staudtMarius:
Resolves #1558
Resolves #1560


src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -183,7 +185,7 @@ public Map<Point, List<ZonedDateTime>> getTimeKeysAfter(ZonedDateTime time) {
           jsonWeatherInputs.stream()
               .map(
                   json -> {
-                    int coordinateId = json.getInt(COORDINATE_ID);
+                    int coordinateId = json.getInt(WEATHER_COORDINATE_ID.toLowerCase());
```
danielfeismann, line 188: is this necessary since it is a string that should be an int?

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactoryTest.groovy
```diff
@@ -193,31 +194,6 @@ class IconTimeBasedWeatherValueFactoryTest extends Specification {
     def model = factory.buildModel(data)
 
     then:
-    !model.equals(expectedResults)
-  }
-
-  def "A IconTimeBasedWeatherValueFactory should throw an Exception if the required field 't2m' is empty"() {
```
danielfeismann, line 199: Are empty fields of required fields covered by other tests?

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
staudtMarius: `WEATHER_COORDINATE_ID` is a string that defines the key for the json map.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactoryTest.groovy
staudtMarius: To check if all needed fields are present, one should use the validation method for sources. In case of csv weather inputs the validation is performed during the read in of the data. For the other weather input types the method `validate` of the source needs to be called explicitly.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
danielfeismann: And it should be an int, which is hard to write in lower case, right?

staudtMarius: JSON is `Map<String, String>`. The method `getInt` uses the string `WEATHER_COORDINATE_ID` to retrieve a string from the map and then converts it to an integer.

`COORDINATE_ID` was the string `coordinateid`. After I replaced it with `FieldNamingStrategy.WEATHER_COORDINATE_ID` the tests were failing, since the json expects a lower case string.

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactoryTest.groovy
danielfeismann: First, do we have a test that tests that this behaviour is still working fine? Second, what do you think, should we add the same validation for the other sources in PSDM as well to have them all the same?

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
danielfeismann: Understood, thanks!

src/test/groovy/edu/ie3/datamodel/io/factory/timeseries/IconTimeBasedWeatherValueFactoryTest.groovy
staudtMarius: Each source should implement a method `validate()`. The idea is that you create a source, call this method and then use the source. One example is the `CsvJointGridContainerSource`.

danielfeismann (approved):

PR #1583, opened 2026-03-24
Use the default time pattern for persisting time series keys

### LINKED ISSUE #766: Use the default time pattern for persisting time series keys instead of the `toString` method

### LINKED ISSUE #884: False Matches from simulation time and weather data (SQL)
Since [SIMONA#574](https://github.com/ie3-institute/simona/issues/574) identified posibilities of false matches from simulation time and weather data (at least) for SQL databases this should be fixed in PSDM.

danielfeismann:
resolves #766
resolves #884 
merge first:
- #1587


src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
```diff
@@ -191,7 +179,7 @@ public Map<Point, List<ZonedDateTime>> getTimeKeysAfter(ZonedDateTime time) {
                         weatherFactory.toZonedDateTime(
                             json.getString(weatherFactory.getTimeFieldString()));
                     if (coordinate.isEmpty()) {
-                      log.warn("Unable to match coordinate ID {} to a point", coordinateId);
+                      logger.warn("Unable to match coordinate ID {} to a point", coordinateId);
```
staudtMarius, line 182: Why do we have a logger in this class and a different one in the base class?

staudtMarius (changes requested):
Looks good to me. I have just one small question.

src/main/java/edu/ie3/datamodel/io/source/couchbase/CouchbaseWeatherSource.java
danielfeismann: I don't know. I just fixed this. However this might be fixed in a separate PR?

staudtMarius: Maybe we could you only one of the loggers. But this could be done in a separate issue.

staudtMarius (approved):

PR #1589, opened 2026-03-25
Handled SonarQube Issues

### LINKED ISSUE #1588: Solve Sonar Issues

danielfeismann:
resolves #1588


src/main/java/edu/ie3/datamodel/io/source/SystemParticipantSource.java
```diff
@@ -34,6 +35,7 @@ public class SystemParticipantSource extends AssetEntitySource {
 
   private static final String THERMAL_STORAGE = "thermalstorage";
   private static final String THERMAL_BUS = "thermalbus";
+  public static final String CONTROLLING_EM = FieldNamingStrategy.CONTROLLING_EM;
```
sebastian-peter: I think what SQ is asking for is just matching against `FieldNamingStrategy.CONTROLLING_EM` in line 66. I think you don't have to declare another static field here.

src/main/java/edu/ie3/datamodel/models/value/WeatherValue.java
```diff
@@ -15,8 +15,7 @@
 import tech.units.indriya.ComparableQuantity;
 
 /** Describes weather as a combination of solar irradiance, temperature and wind values */
-public sealed class WeatherValue implements Value
-    permits WeatherValue.CosmoWeatherValue, WeatherValue.IconWeatherValue {
+public sealed class WeatherValue implements Value {
```
sebastian-peter, line 18: Could you do the same thing for `IdCoordinateInput` while you're at it? :)

sebastian-peter (changes requested):

sebastian-peter (approved):

PR #1596, opened 2026-04-08
Proper handle duplicated headline exceptions in CsvDataSource

### LINKED ISSUE #1595: Proper handle duplicated headline exceptions in `CsvDataSource`
In case two completely identical (also case sensitive) column headlines are handled in `CsvDataSource.buildFieldsToAttributes()` this condition is never reached

https://github.com/ie3-institute/PowerSystemDataModel/blob/d923459322037dd7d3855a63c7db891b6b77c396/src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java#L180-L185

because this fails already in 

https://github.com/ie3-institute/PowerSystemDataModel/blob/d923459322037dd7d3855a63c7db891b6b77c396/src/main/java/edu/ie3/datamodel/io/source/csv/CsvDataSource.java#L173-L178

danielfeismann:
resolves #1595


sebastian-peter (approved):
Thanks for fixing this. I made tiny adaptations, please check if you're fine with them

PR #1619, opened 2026-04-30
false positive tests in csvidcoordinatesourcecosmoit and csvidcoordinatesourceiconit

### LINKED ISSUE #1594: False positive tests in CsvIdCoordinateSourceCosmoIT and CsvIdCoordinateSourceIconIT
The unit tests evaluates to green even when there are case mismatches in the map keys. E.g., expected latGeo vs. actual latgeo

pierrepetersmeier:
Resolves: #1594


pierrepetersmeier:
False positives occurred because containsAll() on map lists ignored case mismatches
(e.g., expected "latgeo" vs. actual "latGeo" from factory).

Changed to strict equality actualStream.data.get().collect(toList()) == expectedStream.collect(toList()),
exposing the mismatch. Updated expected keys to camelCase convention.

"The CsvCoordinateSource is able to create a valid stream from a coordinate file" test
appears redundant with sourceFields + getCoordinate tests. I could be wrong.

danielfeismann (approved):

PR #1641, opened 2026-05-26
Adapting `EnergyBoundariesFlexOptionsResult` to include current SOE

### LINKED ISSUE #1640: Adapt `EnergyBoundariesFlexOptionsResult` to include current SOE
Background described here: https://github.com/ie3-institute/simona/issues/1815

sebastian-peter:
Closes #1640

### DIFF
--- a/src/main/java/edu/ie3/datamodel/models/result/system/EnergyBoundariesFlexOptionsResult.java
+++ b/src/main/java/edu/ie3/datamodel/models/result/system/EnergyBoundariesFlexOptionsResult.java
@@ -15,16 +15,13 @@ import tech.units.indriya.ComparableQuantity;
 /** An energy boundaries flexibility options result of a model (system participant or EM agent). */
 public class EnergyBoundariesFlexOptionsResult extends FlexOptionsResult {
 
-  /**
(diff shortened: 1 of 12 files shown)


src/main/java/edu/ie3/datamodel/models/result/system/EnergyBoundariesFlexOptionsResult.java
```diff
@@ -66,6 +68,8 @@ public String toString() {
         + getTime()
         + ", inputModel="
         + getInputModel()
+        + ", eState="
+        + geteState()
```
staudtMarius: The getter for the energy related quantities could be replaced with the actual field calls.

staudtMarius (changes requested):
Looks good to me. I have just one small suggestion.

src/main/java/edu/ie3/datamodel/models/result/system/EnergyBoundariesFlexOptionsResult.java
sebastian-peter: You mean like this?

staudtMarius (approved):

PR #1662, opened 2026-06-17
Fixed issues regarding determination of additional parameters

### LINKED ISSUE #1661: Problems regarding determination of addtional parameters
There are currently two issues regarding the additional parameters. First, for the grid assets the fields "uuid" and "id" are printed as unsused fields. The second, for uncontrolled ems the field "controlling_em" is present in the addtional parameters.

staudtMarius:
Resolves #1661


src/main/java/edu/ie3/datamodel/io/source/EnergyManagementSource.java
danielfeismann:
```suggestion

// at the start, there are only root ems
```

src/main/java/edu/ie3/datamodel/io/source/EnergyManagementSource.java
```diff
@@ -106,7 +106,11 @@ private static Map<UUID, EmInput> createEmInputs(
                     .map(
                         entityDataTry ->
                             entityDataTry.map(
-                                entityData -> new EmAssetInputEntityData(entityData, null)))
+                                entityData -> {
+                                  // to remove the empty field from the additional data
```
danielfeismann:
```suggestion
                                  // remove the empty field from the additional data
```

danielfeismann (changes requested):

danielfeismann (approved):

PR #1671, opened 2026-06-25
Introduce data model for cable types

### LINKED ISSUE #1669: Introduce new classes for cable data model

danielfeismann:
resolves #1669


src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
```diff
@@ -0,0 +1,445 @@
...
+    this.frequency = frequency;
+    this.skinEffectCoefficient = skinEffectCoefficient;
+    this.proximityEffectCoefficient = proximityEffectCoefficient;
+    this.electricalCapacitance = electricalCapacitance;
+    this.tanDelta = tanDelta;
+    this.circulatingLossFactor = circulatingLossFactor;
+    this.eddyCurrentLossFactor = eddyCurrentLossFactor;
+  }
+
+  @com.fasterxml.jackson.annotation.JsonProperty("core_number")
```
staudtMarius: Is this necessary?

src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
```diff
@@ -0,0 +1,445 @@
...
+      double skinEffectCoefficient,
+      double proximityEffectCoefficient,
+      ComparableQuantity<ElectricCapacitance> electricalCapacitance,
+      double tanDelta,
+      double circulatingLossFactor,
+      double eddyCurrentLossFactor) {
+
+    super(uuid, id);
+
+    Objects.requireNonNull(conductor, "Conductor cannot be null");
```
staudtMarius, line 92: In my opinion, the whole validation should be moved to the validation utils and its subclasses.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
```diff
@@ -0,0 +1,162 @@
...
+   * @param diameter Geometric outer diameter
+   * @param isCompacted Whether the conductor is compacted
+   * @param thermalResistivity Thermal resistivity
+   * @param thermalCapacitance Thermal capacitance
+   * @param areaValue real cross-sectional area
+   * @throws IllegalArgumentException if validation constraints are violated
+   */
+  public ConductorInput {
+    // Validation
+    Objects.requireNonNull(uuid, "Conductor UUID cannot be null");
```
staudtMarius, line 100: Same as for `CableTypeInput`.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
```diff
@@ -0,0 +1,162 @@
...
+    this(
+        uuid,
+        name,
+        material,
+        crossSection,
+        diameter,
+        isCompacted,
+        thermalResistivity,
+        thermalCapacitance,
+        Objects.requireNonNull(area, "Area optional must not be null").orElse(null));
```
staudtMarius, line 81: I think that the `orElse(null)` call can be removed, since the first call will throw an exception if the object is null. If it is not null the later call will not be used.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LayerInput.java
```diff
@@ -0,0 +1,105 @@
...
+   * @param material Material of this layer
+   * @param innerDiameter Inner diameter
+   * @param outerDiameter Outer diameter
+   * @param thermalResistivity Thermal resistivity
+   * @param thermalCapacitance Thermal capacitance
+   * @param area Optional real cross-sectional area
+   * @throws IllegalArgumentException if validation constraints are violated
+   */
+  public LayerInput {
+    // Validation
```
staudtMarius, line 55: Same as the other validation calls.

src/main/java/edu/ie3/datamodel/models/input/connector/CableDeploymentInput.java
```diff
@@ -0,0 +1,85 @@
...
+  /**
+   * Create a new CableDeploymentInput with all required parameters.
+   *
+   * @param layoutFormation Layout formation type.
+   * @param depthCables Laying depth of the cables from ground level to cable center. We keep the
+   *     negative sign for easier integration with Coordinates, thus depthCables must be negative.
+   * @param distanceCables Distance between cables/phases from center to center.
+   * @throws IllegalArgumentException if validation constraints are violated.
+   */
+  public CableDeploymentInput { // Validation
```
staudtMarius, line 35: Same as the other validation calls.

staudtMarius (changes requested):
Here are some remarks from my side. Also, one thing to keep in mind is that with this implementation the cable types themself are not read in by the data model. But the uuid of the cable type is already available to the line type via the additional information map.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
danielfeismann: I don't think so. This would run into a recursion call of the constructor. However, if you know some improvement here I don't see. Please feel free to improve.

src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
danielfeismann: Not necessarily, however snake and camel cases are sometimes used both and I found this as a way to support both when deserialize the inputs. I found a way to have this also for other parameter. But of course we can also try to align to snake case.

src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
danielfeismann: See your point, as we have all of our current validation collected there. However, the way I did here is much shorter, easier to find and in my opinion easier to understand. Or is there another argument I'm currently note aware of.

danielfeismann:
> Here are some remarks from my side. Also, one thing to keep in mind is that with this implementation the cable types themself are not read in by the data model. But the uuid of the cable type is already available to the line type via the additional information map.

Thanks for your effort and the very quick reply @staudtMarius! Yes, I tried to split the PRs so they do not get even bigger than this one already is.

src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
staudtMarius: We had the same discussion for the markov-model PR. There it was decided to move the internal validation to the utils classes.
I would like to keep the model description and the validation in separate locations.

src/main/java/edu/ie3/datamodel/models/input/connector/type/CableTypeInput.java
staudtMarius: I don't think that we could run into troubles. We reading in the data, the field names are always converted into camel case. And internally we only use camel case.
But maybe you see some issues with the current handling.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
```diff
@@ -0,0 +1,116 @@
...
+ * @param name Human-readable id
+ * @param material Material of the conductor (e.g., copper or aluminium)
+ * @param crossSection Real nominal cross-sectional area (electrically effective)
+ * @param diameter Geometric outer diameter of the conductor
+ * @param isCompacted Whether the conductor is compacted
+ * @param thermalResistivity Thermal resistivity of the conductor material
+ * @param thermalCapacitance Thermal capacitance of the conductor material
+ * @param area Optional real cross-sectional area (if different from geometric calculation)
+ */
+public record ConductorInput(
```
staudtMarius: Is there a reason for using records? Also, you could implement the `Uniqueness` interface here and for the other records that contains `UUID`.

staudtMarius (changes requested):
One more question from my side.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
danielfeismann: This safes the getters and equals, etc.. However, if you prefer classes, I don't mind.

staudtMarius: I would prefer to have normal classes here, to be in line with the rest of the data model. But we could also discuss, if we want to change everything to records.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
```diff
@@ -0,0 +1,176 @@
...
+import javax.measure.quantity.Length;
+import org.jspecify.annotations.NonNull;
+import tech.units.indriya.ComparableQuantity;
+
+/**
+ * Represents the conducting core of a cable with its specific geometric and thermal properties.
+ * Unlike {@link LayerInput} layers, the conductor has no inner diameter and includes compaction
+ * information.
+ */
+public class ConductorInput implements InputEntity, Uniqueness, Serializable {
```
staudtMarius: If you rename `name` to `id`, this class could extend `AssetInput`.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LayerInput.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Optional;
+import java.util.UUID;
+import javax.measure.quantity.Area;
+import javax.measure.quantity.Length;
+import tech.units.indriya.ComparableQuantity;
+
+/**
+ * Represents a concentric layer of a cable, including insulation, filler, armor, and jacket layers.
+ */
+public class LayerInput implements InputEntity, Uniqueness, Serializable {
```
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LineTypeInput.java
```diff
@@ -180,7 +180,7 @@ public static final class LineTypeInputCopyBuilder
     private ComparableQuantity<ElectricCurrent> iMax;
     private ComparableQuantity<ElectricPotential> vRated;
 
-    protected LineTypeInputCopyBuilder(LineTypeInput entity) {
+    private LineTypeInputCopyBuilder(LineTypeInput entity) {
```
staudtMarius: Since `LineTypeInput` is not a final class and could therefore be extended, I would keep this constructor protected.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ScreenLayerInput.java
```diff
@@ -0,0 +1,208 @@
...
+import javax.measure.quantity.Area;
+import javax.measure.quantity.Length;
+import org.jspecify.annotations.NonNull;
+import tech.units.indriya.ComparableQuantity;
+
+/**
+ * Represents a cable screen layer with specific parameters for conductor shielding. Extends the
+ * properties of a standard layer with wire-specific parameters.
+ */
+public class ScreenLayerInput implements InputEntity, Uniqueness, Serializable {
```
staudtMarius: Same as above.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LayerInput.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private final UUID uuid;
+  private final String name;
+  private final CableMaterial material;
+  private final ComparableQuantity<Length> innerDiameter;
+  private final ComparableQuantity<Length> outerDiameter;
+  private final ComparableQuantity<ThermalResistivity> thermalResistivity;
+  private final ComparableQuantity<ThermalCapacitance> thermalCapacitance;
+  private final Optional<ComparableQuantity<Area>> area;
```
staudtMarius: Since this field is private, you could remove the optional wrapper here and use `Optional.ofNullable` in the getter. This way the class becomes serializable.

staudtMarius (changes requested):
Here are some more comments from my side.

src/main/java/edu/ie3/datamodel/models/input/connector/type/ConductorInput.java
```diff
@@ -51,40 +50,94 @@ public class ConductorInput implements InputEntity, Uniqueness, Serializable {
...
+      Map<String, String> additionalInformation) {
+    super(uuid, id, operator, operationTime);
     this.material = material;
     this.crossSection = crossSection;
     this.diameter = diameter;
     this.isCompacted = isCompacted;
     this.thermalResistivity = thermalResistivity;
     this.thermalCapacitance = thermalCapacitance;
     this.area = area;
+    if (additionalInformation != null) setAdditionalInformation(additionalInformation);
```
staudtMarius: This check is already included in the set method. Same for the other places.

staudtMarius (changes requested):
One last comment from my side,

staudtMarius (dismissed):

staudtMarius (approved):

PR #1711, opened 2026-08-25
Adding code generation code

### LINKED ISSUE #1710: Add code generation code

staudtMarius:
Resolves #1710


src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
```diff
@@ -1,5 +1,5 @@
 /*
- * © 2021. TU Dortmund University,
+ * © 2026. TU Dortmund University,
```
danielfeismann, line 2: original year should be kept, same for others

src/main/java/edu/ie3/datamodel/io/factory/input/participant/EvcsInputFactory.java
```diff
@@ -71,7 +71,7 @@ protected EvcsInput buildModel(
           e);
     }
 
-    final boolean v2gSupport = data.getBoolean(V2G_SUPPORT);
+    final boolean v2gSupport = data.getBoolean(V_2G_SUPPORT);
```
danielfeismann: this looks odd

src/main/java/edu/ie3/datamodel/io/factory/input/participant/PvInputFactory.java
```diff
@@ -41,8 +41,8 @@ protected PvInput buildModel(
         data.getQuantity(ETA_CONV, StandardUnits.EFFICIENCY);
     final ComparableQuantity<Angle> elevationAngle =
         data.getQuantity(ELEVATION_ANGLE, StandardUnits.SOLAR_ELEVATION_ANGLE);
-    final double kG = data.getDouble(KG);
-    final double kT = data.getDouble(KT);
+    final double kG = data.getDouble(K_G);
```
danielfeismann, line 44: same here

src/main/java/edu/ie3/datamodel/io/factory/result/ConnectorResultFactory.java
```diff
@@ -32,24 +32,26 @@ protected ConnectorResult buildModel(EntityData data) {
 
     UUID inputModel = data.getUUID(INPUT_MODEL);
     ComparableQuantity<ElectricCurrent> iAMag =
-        data.getQuantity(IAMAG, StandardUnits.ELECTRIC_CURRENT_MAGNITUDE);
-    ComparableQuantity<Angle> iAAng = data.getQuantity(IAANG, StandardUnits.ELECTRIC_CURRENT_ANGLE);
+        data.getQuantity(I_A_MAG, StandardUnits.ELECTRIC_CURRENT_MAGNITUDE);
```
danielfeismann, line 35: I_MAG_A and I_MAG_B would be an option to make this a bit nicer?

src/main/java/edu/ie3/datamodel/io/factory/typeinput/SystemParticipantTypeInputFactory.java
```diff
@@ -76,7 +76,7 @@ private SystemParticipantTypeInput buildEvTypeInput(
         data.getQuantity(E_CONS, StandardUnits.ENERGY_PER_DISTANCE);
 
     ComparableQuantity<Power> sRatedDC =
-        data.getQuantity(S_RATED_DC, StandardUnits.ACTIVE_POWER_IN);
+        data.getQuantity(S_RATED_D_C, StandardUnits.ACTIVE_POWER_IN);
```
danielfeismann: also dc requires some solution

src/main/java/edu/ie3/datamodel/models/input/connector/type/LineTypeInput.java
```diff
@@ -120,21 +121,29 @@ public ComparableQuantity<ElectricPotential> getvRated() {
...
-        && x.equals(that.x)
-        && iMax.equals(that.iMax)
-        && vRated.equals(that.vRated);
+    return QuantityUtils.equals(b, that.b)
+        && QuantityUtils.equals(g, that.g)
+        && QuantityUtils.equals(r, that.r)
+        && QuantityUtils.equals(x, that.x)
+        && QuantityUtils.equals(iMax, that.iMax)
+        && QuantityUtils.equals(vRated, that.vRated);
   }
```
danielfeismann, line 148: the way this has been done before is no longer possible?

danielfeismann (changes requested):
Had a first glimpse into this. Need to check still the most of the changes but some first aspects from my side.

src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
staudtMarius: Since th file is newly generated this information is overwritten.

src/main/java/edu/ie3/datamodel/io/factory/result/ConnectorResultFactory.java
staudtMarius: Should we also change the names in the source files? At least in the long run?

danielfeismann: You have to decide this. I_A_MAG is ok with me, I guess the problem is more relevant at other cases where we need to deviate from camel case (e.g. DC, KG, V2G,...)

staudtMarius: I would like to keep it the way it is now. For the other cases (DC, KG, etc.) I will look for a fix.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LineTypeInput.java
staudtMarius: With this method we check if two quantities are equivalent to each other. If not false would be returned for `1000 V == 1 kV`. Before we explicitly converted the quantites to a specific unit. I wanted to remove this.

danielfeismann: Not sure on this, but aren't there cases where we require the right prefix of the unit (e.g. kV) and thus this can cause problems?

staudtMarius: We only need this in SIMONA and for writing the information to the file. The part in SIMONA is not a problem, because we explicitly convert to the correct unit there. As for writing the model to file, I will add an explicit conversion to prevent possible bugs.

src/main/java/edu/ie3/datamodel/io/factory/input/participant/PvInputFactory.java
staudtMarius: The field name is `kG`, therefore, the constants name is `K_G`. For the other fields this converntion was used before.

src/main/java/edu/ie3/datamodel/models/UniqueEntity.java
staudtMarius: Okay, I found a solution for this problem.

src/main/resources/datamodel.yaml
```diff
@@ -0,0 +1,1156 @@
+version: 1.0
```
danielfeismann: we either remove the verison here or may link this automatic to the psdm version

src/main/resources/datamodel.yaml
```diff
@@ -0,0 +1,1156 @@
...
+        - name: geoPosition
+          type: Point
+          description: "The coordinates of this node, especially relevant for geo-dependant systems, that are connected to this node."
+        - name: voltLvl
+          type: VoltageLevel
+          keys: [voltLvl, vRated]
+          description: "Voltage level of this node."
+        - name: subnet
+          type: int
+          description: "Subgrid of this node."
```
danielfeismann: this may be fixed to
```suggestion
          description: "Subgrid number of this node."
```

src/main/java/edu/ie3/datamodel/utils/QuantityUtils.java
```diff
@@ -0,0 +1,40 @@
...
+   */
+  public static <Q extends Quantity<Q>> boolean equals(Quantity<Q> a, Quantity<Q> b) {
+    if (a == b) {
+      return true;
+    }
+
+    if (a != null && b != null) {
+      // convert the second quantity
+      return a.equals(b.to(a.getUnit()));
+    }
```
danielfeismann: if unit a and b are not convertable, i asume that a exception will be thrown. is this what we want, or should we make this robust and simple return false in case they are incompatible?

src/codegen/java/edu/ie3/codegen/MethodGenerator.java
```diff
@@ -0,0 +1,336 @@
...
+import javax.lang.model.element.Modifier;
+
+public class MethodGenerator implements HelperMethods {
+
+  private final boolean result;
+  private final ModelDefinition model;
+  private final GenerationConfig genConfig;
+  private final Map<String, ModelDefinition.ComponentDefinition> allComponents;
+
+  public MethodGenerator(
```
danielfeismann: javaDoc

src/codegen/java/edu/ie3/codegen/GenerationConfig.java
```diff
@@ -0,0 +1,123 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Class that contains the generation configuration. */
+public final class GenerationConfig implements HelperMethods {
+
+  public String year = "";
```
danielfeismann: javaDoc

src/codegen/java/edu/ie3/codegen/ResolverUtils.java
```diff
@@ -0,0 +1,518 @@
...
+
+      return ParameterizedTypeName.get(type.name, genericArguments);
+    }
+
+    return switch (name) {
+      case "bool", "boolean" -> TypeName.BOOLEAN;
+      case "int" -> TypeName.INT;
+      case "double" -> TypeName.DOUBLE;
+      default -> resolveClassName(name);
+    };
```
danielfeismann: isPrimitive inconsistency. HelperMethods.isPrimitive (lines 78-83) returns true for "bool", "int", "double", "float", but this maps bool, int, double. So "float" is treated as primitive by one path and fails resolution in another. equals/hashCode use isPrimitive will generate == for a float field, and the getter/setter will fail to resolve the type. Pick a single source of truth.

src/codegen/java/edu/ie3/codegen/ModelGenerator.java
```diff
@@ -0,0 +1,372 @@
...
+    // check for versions
+    if (!modelFile.version.equals(genConfig.version)) {
+      throw new IllegalStateException("Versions don't match!");
+    } else {
+      System.out.println("Generating models with version: " + modelFile.version);
+    }
+
+    // generate the data model
+    generateAll(
+        modelFile.modelToPackage(), modelFile.flatten(), genConfig.flatten(), outputDirectory);
```
danielfeismann: Models (lines 351-371) uses HashMap and thus loses YAML ordering; the flatten() result iteration order in generateAll (line 68) is therefore non-deterministic, which can produce non-deterministic output file ordering and, worse, may process a child before its parent if the parent has no config and the child references it.

src/codegen/java/edu/ie3/codegen/ModelGenerator.java
```diff
@@ -0,0 +1,372 @@
...
+            // add copyright comment
+            builder.addFileComment(
+                "© $S. TU Dortmund University,\n Institute of Energy Systems, Energy Efficiency and Energy Economics,\n Research group Distribution grid planning and operation",
+                year);
+          }
+
+        } catch (Exception e) {
+          throw new IllegalArgumentException(
+              "Exception while adding copyright comment for model: " + model.name, e);
+        }
```
danielfeismann: Wraps only a trivial String.isBlank() check in a try/catch and re-throws as IllegalArgumentException. There is nothing that can realistically fail there; the try/catch is dead.

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+     - If {@code true} the field is required to build this model (default: true).
+
+   * - nested
+     - boolean
+     - If {@code true} the component is a nested model (default: false).
+
+   * - nullable
+     - boolean
+     - If {@code true} the source can provide null values (default: false).
+
```
danielfeismann: nullable documented but unused. The doc (codegen.md:123-125) says "If true the source can provide null values". In MethodGenerator.getGetters (lines 43-51), the getter logic only checks !component.required && !ResolverUtils.hasDefaultExpression(component.type) — it does not consult component.nullable. The field is read (in HelperMethods.toString, line 46) only to decide the getter-name form, not the null-handling. Either the doc is wrong or the implementation is missing the null-handling branch.

danielfeismann: nullable also missing from equals. The generated equals uses Objects.equals for non-primitive/non-quantity types, which is null-safe, so this is fine — but the getter behaviour is inconsistent with the doc's implication that nullable drives the return type.

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+     - String
+     - The expression. It can be multiple lines, if `\n` is used.
+
+   * - modifiers
+     - List
+     - List of all modifiers.
+
+   * - addReturn
+     - boolean
+     - If {@code true} a return statement is inserted before the {@link #expression}.
```
danielfeismann: maybe at that default is true

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+
+   * - annotation
+     - boolean
+     - If {@code true} a {@link Override} annotation is added.
```
danielfeismann: same here, default is true

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+     - List
+     - List with additional method definitions in a copy builder.
+
+   * - keyMapper
+     - Map
+     - Map: key to expression. This is used when providing keys for a field.
+
+   * - excludeFromMethods
+     - List
+     - List with fields that should be excluded from all methods. This applies to setters, toMap, equals, hashCode and toString.
```
danielfeismann: excludeFromMethods doc says it applies to setters, toMap, equals, hashCode and toString. The implementation honours that everywhere, except the generated copy builder: CopyBuilderGenerator does not consult genConfig.excludeFromMethods at all (see generateCopyBuilderField, generateCopyBuilderSetter, generateCopyBuilderGetter). If the intent is that a field excluded from equals should also be excluded from the copy builder, this is a bug; if not, the doc should say so.

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+
+   * - nonCapitalized
+     - List
+     - List of method that should not start capitalized after `get`.
```
danielfeismann: i guess also after set, and copyBuilderMethods.

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
+# Code generation
```
danielfeismann: generateFieldNamingStrategy is undocumented

src/codegen/java/edu/ie3/codegen/ModelGenerator.java
```diff
@@ -0,0 +1,372 @@
...
+    // read in all information
+    ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+
+    ModelFile modelFile = mapper.readValue(modelsFile.toFile(), ModelFile.class);
+    GenerationConfigFile genConfig =
+        mapper.readValue(generationConfig.toFile(), GenerationConfigFile.class);
+
+    // check for versions
+    if (!modelFile.version.equals(genConfig.version)) {
+      throw new IllegalStateException("Versions don't match!");
```
danielfeismann: may one enhance this by providing both unmatching versions

src/codegen/java/edu/ie3/codegen/MethodGenerator.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    for (ModelDefinition.ComponentDefinition component : model.components) {
+      if (excludeFromMethods(component, genConfig.excludeFromMethods)) {
+        continue;
+      }
+
+      arguments.add(CodeBlock.of("$L", component.name));
+    }
+
```
danielfeismann: equals/hashCode ignore isQuantity for toString. In generateHashCode, quantity fields are passed directly to Objects.hash (fine) but in generateEquals a quantity uses QuantityUtils.equals — the generated hashCode for a Quantity field does not call QuantityUtils.hashCode, so two equal quantities may have different hash codes. If that is the intent, document it; otherwise it is a real bug.

docs/readthedocs/io/codegen.md
```diff
@@ -0,0 +1,379 @@
...
+
+With version 9.2 the data model is defined by a YAML file (`src/main/resources/datamodel.yaml`) that is included in the
```
danielfeismann: imho this should be introduced in a separate version alone with no other changes. Also this may justify to be version 10.0.

src/main/java/edu/ie3/datamodel/models/input/connector/type/LineTypeInput.java
```diff
@@ -163,21 +173,26 @@ public String toString() {
...
-      extends AssetTypeInput.AssetTypeInputCopyBuilder<LineTypeInputCopyBuilder> {
+  @Override
+  public LineTypeInputCopyBuilder copy() {
+    return new LineTypeInputCopyBuilder(this);
+  }
 
+  public static class LineTypeInputCopyBuilder
+      extends AssetTypeInputCopyBuilder<LineTypeInputCopyBuilder> {
     private ComparableQuantity<SpecificConductance> b;
+
```
danielfeismann: not sure if these empty lines can be avoided in general?

src/codegen/resources/generation.yaml
```diff
@@ -0,0 +1,74 @@
+version: 1.0
```
danielfeismann: same here with version.

src/main/java/edu/ie3/datamodel/io/naming/FieldNamingStrategy.java
```diff
@@ -6,265 +6,863 @@
 package edu.ie3.datamodel.io.naming;
 
 import edu.ie3.datamodel.models.UniqueEntity;
+import edu.ie3.datamodel.models.input.AssetInput;
+import edu.ie3.datamodel.models.input.AssetTypeInput;
+import edu.ie3.datamodel.models.input.EmInput;
+import edu.ie3.datamodel.models.input.IdCoordinateInput;
+import edu.ie3.datamodel.models.input.MeasurementUnitInput;
```
danielfeismann: imho these imports should be optimized

src/main/java/edu/ie3/datamodel/io/naming/FieldNamingStrategy.java
```diff
@@ -6,265 +6,863 @@
...
+import edu.ie3.datamodel.models.result.thermal.ThermalSinkResult;
+import edu.ie3.datamodel.models.result.thermal.ThermalStorageResult;
+import edu.ie3.datamodel.models.result.thermal.ThermalUnitResult;
+import edu.ie3.datamodel.utils.CollectionUtils;
 
-/** Class that contains all known field names. */
-public class FieldNamingStrategy {
+public class FieldNamingStrategy extends FieldNamingStrategyAdditions {
+  public static final String ACTIVE_POWER_GRADIENT = "activePowerGradient";
+
```
danielfeismann: same here if the empty lines can be avoided

src/main/java/edu/ie3/datamodel/io/naming/FieldNamingStrategyAdditions.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * © 2026. TU Dortmund University,
+ * Institute of Energy Systems, Energy Efficiency and Energy Economics,
+ * Research group Distribution grid planning and operation
+*/
+package edu.ie3.datamodel.io.naming;
+
+public class FieldNamingStrategyAdditions {
```
danielfeismann: Sonar:
Check warning: src/main/java/edu/ie3/datamodel/io/naming/FieldNamingStrategyAdditions.java#L8
Add a private constructor to hide the implicit public one.

build.gradle
```diff
@@ -105,6 +106,11 @@ dependencies {
   implementation 'commons-io:commons-io:2.22.0' // I/O functionalities
   implementation 'commons-codec:commons-codec:1.22.1' // needed by commons-compress
   implementation 'org.apache.commons:commons-compress:1.28.0' // I/O functionalities
+
+  // model generation
+  codegenImplementation 'com.palantir.javapoet:javapoet:0.19.0'
+  codegenImplementation 'tools.jackson.dataformat:jackson-dataformat-yaml:3.2.2'
+  codegenImplementation 'com.github.ie3-institute:PowerSystemUtils:3.2.2'
```
danielfeismann: utils are already part of this

build.gradle
```diff
@@ -105,6 +106,11 @@ dependencies {
   implementation 'commons-io:commons-io:2.22.0' // I/O functionalities
   implementation 'commons-codec:commons-codec:1.22.1' // needed by commons-compress
   implementation 'org.apache.commons:commons-compress:1.28.0' // I/O functionalities
+
+  // model generation
+  codegenImplementation 'com.palantir.javapoet:javapoet:0.19.0'
```
danielfeismann: is there a javapoet that is not by palantir? :)

danielfeismann (changes requested):
Some aspects by Joe and also by myself. Also the question if the code generation requires tests of these methods. Furthermore I would like to propose to introduce this in a separate release e.g. 10.0.

PR #1731, opened 2026-09-08
fix overshadowed ParsingException in system participant factories

### LINKED ISSUE #1722: Overshadowing of errors
In the following code, if an exception is thrown, this exception is overshadowed by the error that is thrown on the second `data.getField(Q_CHARACTERISTICS)` call. The cause is that the method removes the key.

```java
    ReactivePowerCharacteristic qCharacteristics;
    try {
      qCharacteristics = ReactivePowerCharacteristic.parse(data.getField(Q_CHARACTERISTICS));
    } catch (ParsingException e) {
      throw new FactoryException(
          "Cannot parse the following reactive power characteristic: '"
              + data.getField(Q_CHARACTERISTICS)
              + "'",
          e);
    }
```

One solution would be to save the value as a variable before. There are also other factories with the same issue.

uhlmanngina:
Fixes #1722

`FactoryData.getField()` removes the entry from the field map. Calling it a
second time for the same key inside a catch block (to build the error
message) therefore failed with "Field not found", overshadowing the actual
ParsingException. Field values are now read once before the try block.


staudtMarius:
@uhlmanngina could you add a Message to the CHANGELOG file?

uhlmanngina:
Done, thanks for the hint!

staudtMarius (approved):
Looks good to me. Thanks for taking care of the issue.

